PR #30, opened 2019-07-27
Added confirm for submitSponsorTimes

OfficialNoob:


content.js
```diff
@@ -670,4 +671,4 @@ function getYouTubeVideoID(url) { // Returns with video id else returns false
   var regExp = /^.*((youtu.be\/)|(v\/)|(\/u\/\w\/)|(embed\/)|(watch\?))\??v?=?([^#\&\?]*).*/;
   var match = url.match(regExp);
   return (match && match[7].length == 11) ? match[7] : false;
-}
\ No newline at end of file
+}
```
ajayyy: What's up with this? Was there a missing bracket?

ajayyy: What's up with this? Was there a missing bracket?

ajayyy: Oh, you added a new line

PR #32, opened 2019-07-27
Added references

OfficialNoob:


ajayyy:
Does this just make it so `document.getElementById` can be called less? If so, that's nice.

PR #33, opened 2019-07-27
Submit Hotkey

OfficialNoob:


content.js
```diff
@@ -98,22 +102,21 @@ chrome.runtime.onMessage.addListener( // Detect URL Changes
...
+    if (document.activeElement === video) {
+        if (e.code == hotkeys.submit) {
+            if (hasData) { // Submit!!!
+                submitSponsorTimes();
+                hasData = false;
+            } else { // First time
+                startSponsorClicked();
+                hasData = true;
+            }
+        }
```
ajayyy: This won't work, as clicking once indicates the start of a sponsor, then clicking again should indicate the end. I think a separate button (I used single quote) should be used for the submitting hot key. Someone might want to submit multiple at once.

ajayyy:
After looking at it again, I can see how it can work, but I don't really like doing it this way, I feel it's easier to make mistakes. I think it's best to find all the sponsors in the video, THEN submit. This makes it so the user is sure those are the correct sponsors.

If you do want it this way for yourself, you can make it an option, but I don't want that to be the default.

OfficialNoob:
Submitting one at a time has the benefit of being able to cancel per sponsor

content.js
OfficialNoob: Code works however needs change

ajayyy:
That could be resolved with a UI change.

And sadly, you can't just use a variable created there because it has to sync up with the popup. I recommend checking if the document.getElementById("submitButton").style != "none".

I'll merge it in in if you make it an option and not on by default.

OfficialNoob:
I dont like the global popup as its not per video

content.js
```diff
@@ -107,13 +112,21 @@ document.onkeydown = function(e) {
     let video = document.getElementById("movie_player");
     //is the video in focus, otherwise they could be typing a comment
     if (document.activeElement === video) {
-        if (e.code == hotkeys.submit) {
-            if (hasData) { // Submit!!!
+        if (e.code == Settings.hotkeys.submit) {
+            if(Settings.onekey) {
+              if (hasData) { // Submit!!!
```
ajayyy: as I've said before,  this is not going to work sadly. The data can be changed by clicking on the buttons as well as hotkeys. I'd recommend checking if the submit button is visible instead of using that variable.

content.js
```diff
@@ -2,8 +2,13 @@ if(id = getYouTubeVideoID(document.URL)){ // Direct Links
...
 }
 
-var hotkeys = {};
-hotkeys.submit = "Semicolon";
+var Settings = {} // Needs to be Synced with global settings
+
+Settings.hotkeys = {}; // TEMP
+Settings.onekey = false; // TEMP
+Settings.hotkeys.submit = "Quote"; // TEMP
+Settings.hotkeys.start = "Semicolon"; // TEMP
```
ajayyy: Can you set this up so the settings save? Right now it does nothing as it is by default just the same

content.js
```diff
@@ -107,13 +112,21 @@ document.onkeydown = function(e) {
...
                 hasData = false;
-            } else { // First time
+              } else { // First time
                 startSponsorClicked();
                 hasData = true;
+              }
+            }else{
+              if(e.code == Settings.hotkeys.start){
+                startSponsorClicked();
+              } else if (e.code == Settings.hotkeys.submit) {
```
ajayyy: code is a number not a string, right?

OfficialNoob: code is a string

content.js
OfficialNoob: You have a good point there last time I did not change it because I was hoping the global popup would be changed to settings

ajayyy: There are also buttons on the player

OfficialNoob: Is this fixed now?

content.js
```diff
@@ -114,7 +112,7 @@ document.onkeydown = function(e) {
     if (document.activeElement === video) {
         if (e.code == Settings.hotkeys.submit) {
             if(Settings.onekey) {
-              if (hasData) { // Submit!!!
+              if (document.getElementById("submitButton").style != "none") { // Submit!!!
                 submitSponsorTimes();
                 hasData = false;
```
ajayyy: remove this

content.js
OfficialNoob: It would be there no place to change them or to store the default values

content.js
OfficialNoob: Done :D

content.js
ajayyy: should be, you'd have to test it

content.js
ajayyy: Yea, for now you can put it in the options section of the popup

OfficialNoob: There is in the option page currently???

ajayyy: Yes there is, click the options button  the popup

OfficialNoob: Waiting for the UI changes

ajayyy:
https://github.com/ajayyy/SponsorBlock/pull/163

PR #49, opened 2019-07-30
Old function in-use

OfficialNoob:


ajayyy:
These duplicate functions really need to be moved to a separate js file at some point.

PR #73, opened 2019-08-04
Made UUID less predictable

OfficialNoob:


ajayyy:
Maybe something else can be used as the seed on the non secure browsers. Which are considered not secure?

OfficialNoob:
Well the best way would be to generate it serverside

ajayyy:
Seems to lower security

ajayyy:
User id is a password, it should be decided by the client

OfficialNoob:
Its more like a token then a password because there is "no" user input.
In order to enforce a standard and prevent attacks:
server side generation would be easier and the token can change every x amount of time
https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues
I think this is safe as the value is being sent to and validated by the same server anyway

OfficialNoob:
@ajayyy why is this not ready? it should result in the same length value even on older devices

ajayyy:
These aren't in the UUID format yet

ajayyy:
Maybe https://www.uuidgenerator.net/api can be used?

OfficialNoob:
@ajayyy why would you trust uuidgenerator.net and not your own server?

OfficialNoob:
Why are you using a UUID anyway?
"Do not assume that UUIDs are hard to guess"
https://tools.ietf.org/html/rfc4122

ajayyy:
That would just be to make things easier.

ajayyy:
Okay, I see now. That makes sense then, I'll switch to your system then.

ajayyy:
@OfficialNoob Make sure you don't use `var` when you don't have to.

PR #74, opened 2019-08-04
Options "inprovments" [INDEV]

OfficialNoob:


ajayyy:
This isn't going to work since if you target all body tags, then it will break as the info menu on the YouTube site

PR #72, opened 2019-08-04
Added Utils file + Updated Parser

OfficialNoob:
This is my version of the Utils
There is also https://github.com/ajayyy/SponsorBlock/pull/85

PR #80, opened 2019-08-05
Improve Youtube parsing function

Jackymancs4:
Hello @ajayyy 
I've been following your work on the project since I saw it on Reddit some days ago. Really well done!

For reasons I don't remember I hit a link with this format https://www.youtube.com/watch?time_continue=x&v=xxxxxxxxxxx and the popup didn't recognize it as a valid youtube url.

So, I rewrote the function that parse the url to make it more resilient, and IMO clear. (It took me awhile to decrypt the regex).

Since the function is copy pasted in three places I was planning to do the very same thing of #72 (before it was opened), but after seeing it I choose for this easy-to-merge PR.

If you prefer and are interested in the PR, I can still make an `util.js` here and @OfficialNoob can move the other needed functions there afterwards (rebasing #72 on top of this)

### DIFF
--- a/background.js
+++ b/background.js
@@ -263,10 +263,25 @@ function sendRequestToServer(type, address, callback) {
   xmlhttp.send();
 }
 
-function getYouTubeVideoID(url) { // Return video id or false
-  var regExp = /^.*((youtu.be\/)|(v\/)|(\/u\/\w\/)|(embed\/)|(watch\?))\??v?=?([^#\&\?]*).*/;
-  var match = url.match(regExp);
-  return (match && match[7].length == 11) ? match[7] : false;
+// Returns with video id else returns false
+function getYouTubeVideoID(url) {
+  let urlObject = new URL(url)
+
+  if (urlObject.host != "www.youtube.com") {
+      return false
+  }
+
+  if (urlObject.pathname == "/watch/" || urlObject.pathname == "/watch") {
+      let id = urlObject.searchParams.get("v");
+      return id.length == 11 ? id : false
+  }
+
+  if (urlObject.pathname.startsWith("/embed/")) {
+      let id = urlObject.pathname.slice("/embed/".length)
+      return id.length == 11 ? id : false
+  }
+
+  return false
 }
 
 //uuid generator function from https://gist.github.com/jed/982883
--- a/content.js
+++ b/content.js
@@ -1144,16 +1144,25 @@ function sendRequestToCustomServer(type, fullAddress, callback) {
   xmlhttp.send();
 }
 
-function getYouTubeVideoID(url) { // Returns with video id else returns false
-  var regExp = /^.*((youtu.be\/)|(v\/)|(\/u\/\w\/)|(embed\/)|(watch\?))\??v?=?([^#\&\?]*).*/;
-  var match = url.match(regExp);
-  var id = new URL(url).searchParams.get("v");
-  if (url.includes("/embed/")) {
-    //it is an embed, don't search for v
-    id = match[7];
+// Returns with video id else returns false
+function getYouTubeVideoID(url) {
+  let urlObject = new URL(url)
+
+  if (urlObject.host != "www.youtube.com") {
+      return false
   }
 
-  return (match && match[7].length == 11) ? id : false;
+  if (urlObject.pathname == "/watch/" || urlObject.pathname == "/watch") {
+      let id = urlObject.searchParams.get("v");
+      return id.length == 11 ? id : false
+  }
+
+  if (urlObject.pathname.startsWith("/embed/")) {
+      let id = urlObject.pathname.slice("/embed/".length)
+      return id.length == 11 ? id : false
+  }
+
+  return false
 }
 
 //returns the start time of the video if there was one specified (ex. ?t=5s)
--- a/popup.js
+++ b/popup.js
@@ -1127,11 +1127,25 @@ function runThePopup() {
     xmlhttp.send();
   }
   
-  function getYouTubeVideoID(url) { // Returns with video id else returns false
-    var regExp = /^.*((youtu.be\/)|(v\/)|(\/u\/\w\/)|(embed\/)|(watch\?))\??v?=?([^#\&\?]*).*/;
-    var match = url.match(regExp);
-    var id = new URL(url).searchParams.get("v");
-    return (match && match[7].length == 11) ? id : false;
+  // Returns with video id else returns false
+  function getYouTubeVideoID(url) {
+    let urlObject = new URL(url)
+
+    if (urlObject.host != "www.youtube.com") {
+        return false
+    }
+
+    if (urlObject.pathname == "/watch/" || urlObject.pathname == "/watch") {
+        let id = urlObject.searchParams.get("v");
+        return id.length == 11 ? id : false
+    }
+
+    if (urlObject.pathname.startsWith("/embed/")) {
+        let id = urlObject.pathname.slice("/embed/".length)
+        return id.length == 11 ? id : false
+    }
+
+    return false
   }
   
 //end of function


ajayyy:
Ooh, I like this. Can you make the host allow *.YouTube.com though?

Does the getVideoStartTime still work?

Jackymancs4:
> Can you make the host allow *.YouTube.com though?

I assume you refer to this line
```if (urlObject.host != "www.youtube.com") {```

While yes, I wonder why? Every `*.YouTube.com` always redirected to `www.YouTube.com` in my tests so redirecting to it seems safer to me.

> Does the getVideoStartTime still work?

My function behaves exactly as before except for returning the id also for https://www.youtube.com/watch?time_continue=x&v=xxxxxxxxxxx and getYouTubeVideoStartTime seems to natively handle `time_continue`, so nothing is changed in this regard.

**Comment by OfficialNoob** [comment]
Fixed my parser :D

**Comment by Jackymancs4** [comment]
I disagree. While you fixed the specific bug, you removed `match[7].length == 11` which is an important sanitization check.
Now `youtube.com/watch?v=badurl` and `youtube.com/embed/badurl` passes, which may (or may not) cause problems to the server.

My opinion is still that you or @ajayyy should use my version.

Regular expression can be tricked in non trivial ways, while relying on the browser url parser (`new URL()`) is the more solid and scalable approach from both readability and reliability point of view.

Since you already use `new URL()` for the params, the initial regex check is also an unnecessary performance overhead (regex are heavy tasks).

All of this is just my technical opinion, so no offense intended here ;)

**Comment by OfficialNoob** [comment]
Ok I added the check back and it fixes still the problem

**Comment by OfficialNoob** [comment]
If you dont want to use Regex what about this?https://gist.github.com/OfficialNoob/3012f5b7884ca7ae27b5ada4fc601c00

**Comment by ajayyy** [comment]
I like this PR's solution as it is a lot more readable

**Comment by Jackymancs4** [comment]
I put up a little test here 
https://gist.github.com/Jackymancs4/0d509290f35c926994b6417e909e72c3
where I compare every proposed function (with references).

It's pretty rought but it show my point.

Only my solution works only on youtube and only on valid urls.
Your last two proposition matched `https://www.notyoutube.com/watch?v=xxxxxxxxxxx` and `https://www.notyoutube.com/embed/xxxxxxxxxxx` (think of scammyyoutube or any other not youtube domain), which seems pretty unsafe to me.

**Comment by OfficialNoob** [comment]
The extension should only have permission for *.youtube.com

**Comment by OfficialNoob** [comment]
For your PR you should add a alert when its running on a different domain like "Disable Extension now!!!!" if it can run on notyoutube.com there are more problems 😒

**Comment by Jackymancs4** [comment]
Tecnically it also have full priviledge for https://github.com/ajayyy/SponsorBlock/blob/18909ffef6577d9ec7178f860d2529fe6414cdda/manifest.json#L43

And BTW the popup is not restricted by those permissions, as it needs only the tabs permission, granted here:
https://github.com/ajayyy/SponsorBlock/blob/18909ffef6577d9ec7178f860d2529fe6414cdda/manifest.json#L40 

> For your PR you should add a alert when its running on a different domain like "Disable Extension now!!!!" if it can run on notyoutube.com there are more problems 

What?

**Comment by OfficialNoob** [comment]
So the problem is https://sponsor.ajay.app/* would make a fake video?

**Comment by OfficialNoob** [comment]
Does the tabs permission give more control then is needed?

**Comment by Jackymancs4** [comment]
1. You were wrong about this sentence: `The extension should only have permission for *.youtube.com`, so I pointed it out. No other assumptions have been made on my side.

2. According various issues www.youtube-nocookies.com and invidious domains may (or may not) be added, so strictly relying on the permission seems a little naive.

3. the popup is **not** restricted by those permission, it may (or may not) detect https://www.notyoutube.com/embed/xxxxxxxxxxx as good and valid youtube video, and **this** may (or may not) cause problems to the extension (I have not thoroughly investigated how this is handled).

Regardless, I hope you'll agree that a more secure approach is always better than a less secure one.

> Does the tabs permission give more control then is needed?

> I have not thoroughly investigated how this is handled

**Comment by OfficialNoob** [comment]
1. It SHOULD only have access to www.youtube.com (Does not mean it will)
2. Even if more domains get added the any of the content script should not be ran not just this one function
3. The content script is should only run on website it has permission for am I wrong?
Tabs. Hopefully im right or this is a problem

**Comment by Jackymancs4** [comment]
1. Obvioulsy not because it SHOULD only have access to www.youtube.com and the address of the server for API consumption, or more than half of the extension falls apart.
2. Yes, BUT the very specific function we are discussing is shared between content script AND the popup ( + background) which follow different rules that MUST be kept into consideration.
3. You are right.

**Comment by OfficialNoob** [comment]
0. This is fun lol
1. I meant it should only have access to youtube.com at the DOM level the API/s dont need that access all they need is a CORS Header
3. Oh good

**Comment by OfficialNoob** [comment]
@Jackymancs4 you made a script to call my functions "Terrible" Instead of me just add a check
sorry I did not mean to be defensive and a single point of failure in security is bad and I should not have suggested that such check was not needed

**Comment by Jackymancs4** [comment]
> @Jackymancs4 you made a script to call my functions "Terrible"

This is definetely not what I meant. The idea was that it was `Nice` if the test passed and `Terrible` if it did not but this in regard of any function, not specifically yours (which consistently passed the most important cases). It seemed funny in my head, but I'm so sorry to have conveyed the wrong message!

Please keep in mind that, while I try my best, English is not my first language and I may fail to fully express myself!

I only want the best out of this, and you did an awesome job as a contributor for this project, so thank you!

Regarding gist.github.com/OfficialNoob/2026f17032b1f1cf9b3d2d8bf4f2c558
the problem is that `window.location.host` and most likely the whole `window` object (not sure though) are only available on the content script.

If you share that function with the `util.js` file it will work on the content script and fail on both the popup and background (either with an error or silently).

My opinion is to just keep the function as an arbitrary validation function, which take as an input a string (the url) and is called differently in different locations.

Where I live is pretty late, so, if needed I will comment further tomorrow.

**Comment by OfficialNoob** [comment]
Ok I changed it to the taking an input way

**Comment by OfficialNoob** [comment]
https://github.com/OfficialNoob/SponsorBlock/commit/02e11503cba1ee420a338e200e1a873ff71d081e

**Comment by ajayyy** [comment]
I do agree that one function should handle all parsing in the future
(invidious, etc.)

About permissions, only Firefox seems to restrict the extension in my
testing.

**Comment by OfficialNoob** [comment]
Why do you need invidious cant you get the data from the client anyway
Server side validation?

**Comment by OfficialNoob** [comment]
I did a test myself I put a alert at the top of the content script and it did not run on example.com

**Comment by Jackymancs4** [comment]
> What ever function gets used I think it should go in utils.js

Completely agree.

Besides, I updated https://gist.github.com/Jackymancs4/0d509290f35c926994b6417e909e72c3 to take into account some more edge cases.

Your last proposal (getYouTubeVideoID9) taken from https://github.com/OfficialNoob/SponsorBlock/commit/02e11503cba1ee420a338e200e1a873ff71d081e almost never works because of `(obj.host !== "www.youtube.com" || "www.youtube-nocookie.com")` being equal to `(obj.host !== "www.youtube.com" || true)`

In my gist, getYouTubeVideoID7 is a fixed version of your function.

Still, it is my opinion that getYouTubeVideoID8 should be used as it handle both all the edge cases and address the readability concerns of @ajayyy . I will soon update this PR to reflect this.

Also, regarding supporting invidious, I think it is outside the scope of this PR and any discussion should continue on the relative issue.

**Comment by OfficialNoob** [comment]
@Jackymancs4 I think I made my function better could you check? https://github.com/OfficialNoob/SponsorBlock/commit/c6c8d7de491281e38d21c0d9124bdb0e6c1d7644

**Comment by Jackymancs4** [comment]
Unluckily, it does not work at all, because `let` is scope declaration so obj will die as soon that is leaves the try statement. You should use var or declare an empty obj before the try.

I see that you fixed this:
```
        return false
        console.error("[SB] Unable to parser URL");
```

And it still fails `https://www.youtube.com/watch/?v=xxxxxxxxxxx` and `https://www.youtube.com/embed/wJuWuJBbc1s/` which are valid cases.

Besides I love this: `if(!["www.youtube.com","www.youtube-nocookie.com"].includes(obj.host)) return false` !

Jackymancs4:
I updated https://gist.github.com/Jackymancs4/0d509290f35c926994b6417e909e72c3 with `getYouTubeVideoID11` being your function from https://github.com/OfficialNoob/SponsorBlock/commit/c6c8d7de491281e38d21c0d9124bdb0e6c1d7644 and `getYouTubeVideoID10` the same shipping the fixes.

OfficialNoob:
As I keep messing it up I made a unittest
https://github.com/ajayyy/SponsorBlock/pull/84

OfficialNoob:
I just tested it and both https://www.youtube.com/watch/?v=xxxxxxxxxxx and https://www.youtube.com/embed/wJuWuJBbc1s/ work

ajayyy:
By invidious support, I did not mean using the invidious API, I meant working on the invidious site. The invidious API should be replaced with getting data from the client at some point as there is no need to waste the invidious's APIs resource

ajayyy:
Why two functions?

OfficialNoob:
I have added a lot of fail safes to my parser now lol
https://github.com/OfficialNoob/SponsorBlock/commit/b964d93ea948b94515a7a93dc573c5ab5a52be3c

OfficialNoob:
Can this PR be closed?

PR #84, opened 2019-08-07
Created a unittest

OfficialNoob:


Jackymancs4:
Nice! How do I manually run? (even if POC)

Jackymancs4:
Also, I suggest to replace all `wJuWuJBbc1s` with a generic `xxxxxxxxxxx` as it should make no difference.

ajayyy:
Nice!

OfficialNoob:
@Jackymancs4 maybe put it in a html file with <script> tags then in the console run ALL()

ajayyy:
Can you change these to camelCase?

OfficialNoob:
@ajayyy enforce with Linter in free CI xD

OfficialNoob:
@ajayyy Is this project getting unit tests and Linters?
[Github Actions](https://github.com/features/actions)

ajayyy:
It should. I think it would be most useful on the server code. Especially if I change each API endpoint to be a callable function, then it would be easy to make some tests.

OfficialNoob:
:D maybe CI will be added in 2020

ajayyy:
Valve time

PR #89, opened 2019-08-08
Refactor: make background non-persistent

### LINKED ISSUE #88: Use non-persistent background page
The persistent background page is loaded persistently, so even when there are no events to handle and even no YouTube videos (tabs or embedded frames) are open. This is wasteful since most of the time user does not watch videos let alone constantly actively interact with the server.

[How to do it.](https://developer.chrome.com/extensions/background_migration)

I think this refactoring might simplify #87.

bershanskiy:
Fixes #88.

This PR simply moves all the background.js local variables into browser-provided storage. Variables include:
- `serverAddress` (from config.js)
- `previousVideoID`
- `userID`
- `sponsorVideoID`

Also, while at it I refactored two things:
- removed `shownInstallPage` variable because it's redundant: it is always unset if on the first install when `userID` is unset.
- In the manifest, removed `"help/index.html"` and `"help/style.css"` from `web_accessible_resources` because the help page is shown in a separate tab aka in extension's own context and therefore they do not have to be accessible.


background.js
```diff
@@ -183,71 +187,53 @@ function videoIDChange(currentVideoID, tabId) {
...
+
+    chrome.storage.sync.set({
+      "sponsorVideoID": currentVideoID
     });
-  } else {
-    previousVideoID = currentVideoID;
-  }
-}
 
-function getUserID(callback) {
```
bershanskiy: I deleted `getUserID` because there is no real need for it: the id is generated and saved right during the first install before any YouTube video is loaded (aka before the content script is injected and can try to request it).

ajayyy:
Great, thanks for this! The showInstallPage variable is necessary however as it syncs per user. If it is not used, the install page appears every time it is installed, including a second install. It also appears every time you sign into a school/workplace Chromebook (it reinstalls extensions). The on install is also called every time the extension is reloading, so the help page would constantly appear while developing.

So, if you could add that back, that would be great!

Thanks for the PR, I'll look into merging it when I get back home this weekend :)

ajayyy:
The userid should also not be reset after every install.

bershanskiy:
@ajayyy I can add `showInstallPage` back, but I'm not sure you need it. You can avoid unnecessary complexity if you generate `userID` on the first install and then check its existence to know if it is the first install (aka `showInstallPage` is true if and only if `userID` does not exist).

As is, this will open popup only for users installing for the first time (desired) or updating without `userID` set (aka ones who never voted nor submitted skip times), which is undesired.

I can fix the undesired help opening with a check for `showInstallPage` and delete it. Then (after most people install the patched version) we can remove logic referencing `showInstallPage`.

I can add `showInstallPage` back, but I don't really see the point: for most server calls, you need to look up `serverAddress` and `userID` and you can do that in one call. Otherwise, you need to use `getUserID` which checks for the existence of `userID` (one storage call) and then look up other information (e.g., `serverAddress` with one more storage call).

Should I not generate `userID` on install? Should I add `showInstallPage` back just for the sake of it?

ajayyy:
Oh, sorry. I seem to have missed that when reading the code. That sounds good, no need to save two variables when unnecessary.

Maybe it would be a good idea to check for showInstallPage for now and delete it.

background.js
```diff
@@ -62,14 +54,39 @@ chrome.runtime.onMessage.addListener(function (request, sender, callback) {
...
-      //save that this happened
-      chrome.storage.sync.set({shownInstallPage: true});
+  // TODO (shownInstallPage): remove shownInstallPage logic after sufficient amount of time,
+  // so that people have time to upgrade and move to shownInstallPage-free code.
+  chrome.storage.sync.get(["userID", "shownInstallPage"], function(result) {
+    const userID = result.userID;
+    // TODO (shownInstallPage): delete row below
+    const shownInstallPage = result.shownInstallPage;
+
+    // If there is no userID, then it is the first install.
```
bershanskiy: User id is not reset if it was already generated.

bershanskiy:
> The userid should also not be reset after every install.

It is not reset. See the line I commented on (on GitHub)

> I like the idea of checking for showInstallPage for now and delete it.

I just pushed a commit addressing that. Also, I added comments describing exact future changes that would remove any mention of `shownInstallPage`.

ajayyy:
Merge second

ajayyy:
Looks like this is missing a few changes needed in content.js, popup.js and config.js.example. I'll add those now.

ajayyy:
Oh, the issue is that onInstallStuff isn't done when showInstallPage isn't true, I'll change that

background.js
```diff
@@ -136,16 +156,17 @@ function submitVote(type, UUID, callback) {
 function submitTimes(videoID, callback) {
   //get the video times from storage
   let sponsorTimeKey = 'sponsorTimes' + videoID;
-  chrome.storage.sync.get([sponsorTimeKey], function(result) {
+  chrome.storage.sync.get([sponsorTimeKey, "serverAddress"], function(result) {
```
ajayyy: You forgot to add userID here, so `undefined` was submitted instead.

ajayyy:
@bershanskiy 
image

Is this not supported in Firefox?

background.js
bershanskiy: I think we'd benefit from some test cases :)

bershanskiy:
@ajayyy Yes, [Firefox does not support non-persistent background pages](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/background). This means the pages will be persistent, but everything will function as normal.

PR #90, opened 2019-08-08
Removed tabs permission

OfficialNoob:
Removing use of tab.url as its not needed
Status: Awaiting Review

### DIFF
--- a/background.js
+++ b/background.js
@@ -5,9 +5,9 @@ var userID = null;
 var sponsorVideoID = null;
 
 chrome.tabs.onUpdated.addListener(function(tabId, changeInfo, tab) {
-  chrome.runtime.sendMessage({
-    message: "update",
-  });
+	chrome.tabs.sendMessage(tabId, {
+		message: 'update',
+	});
 });
 
 chrome.runtime.onMessage.addListener(function (request, sender, callback) {
--- a/background.js
+++ b/background.js
@@ -29,7 +29,14 @@ chrome.runtime.onMessage.addListener(function (request, sender, callback) {
 
     //this allows the callback to be called later
     return true;
-  }
+  } else if (request.message == "alertPrevious") {
+	  chrome.notifications.create("stillThere" + Math.random(), {
+		  type: "basic",
+		  title: "Do you want to submit the sponsor times for video id " + request.previousVideoID + "?",
+		  message: "You seem to have left some sponsor times unsubmitted. Go back to that page to submit them (they are not deleted).",
+		  iconUrl: "./icons/LogoSponsorBlocker256px.png"
+	  });
+  }	
 });
 
 //add help page on install
--- a/background.js
+++ b/background.js
@@ -5,7 +5,7 @@ var userID = null;
 chrome.tabs.onActivated.addListener(
   function(activeInfo) {
     chrome.tabs.get(activeInfo.tabId, function(tab) {
-        TabUpdate(activeInfo.tabId);
+        tabUpdate(activeInfo.tabId);
     })
   }
 );
@@ -15,7 +15,7 @@ chrome.tabs.onUpdated.addListener(function(tabId, changeInfo, tab) {
     if (changeInfo != undefined) {
       //if URL changed
       if (tabId) { 
-        TabUpdate(tabId);
+        tabUpdate(tabId);
       }
     }
 });
@@ -164,7 +164,7 @@ function submitTimes(videoID, callback) {
   });
 }
 
-function TabUpdate(tabId) {
+function tabUpdate(tabId) {
   //send a message to the content script
   chrome.tabs.sendMessage(tabId, {
     message: 'update'
--- a/background.js
+++ b/background.js
@@ -4,6 +4,12 @@ var userID = null;
 //the last video id loaded, to make sure it is a video id change
 var sponsorVideoID = null;
 
+chrome.tabs.onUpdated.addListener(function(tabId, changeInfo, tab) {
+  chrome.runtime.sendMessage({
+    message: "update",
+  });
+});
+
 chrome.runtime.onMessage.addListener(function (request, sender, callback) {
   if (request.message == "submitTimes") {
     submitTimes(request.videoID, callback);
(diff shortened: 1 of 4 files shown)


ajayyy:
Can you use camelCase for function and variable names. (First letter lower case)

OfficialNoob:
@ajayyy sure such rules can also be enforced with a linter

OfficialNoob:
@ajayyy it should be done now it toke longer then expected because I tried onhashchange instead of tabUpdate did not work

ajayyy:
`tabUpdate` did work?

OfficialNoob:
Think so but i removed the function as all it needed to do was message the activeTab on chrome.tabs.onUpdated what I tried todo was detect the url change directly from the contentscript using onhashchanged witch did not work

ajayyy:
I don't think that'll work since it has to be triggered by a user action

OfficialNoob:
@ajayyy what will not work? this PR does not need tabs or activeTab permissions
by "activeTab " I meant the activeTab not the permission

OfficialNoob:
"You can use most chrome.tabs methods and events without declaring any permissions in the extension's manifest file. However, if you require access to the url, title, or favIconUrl properties of tabs.Tab, you must declare the "tabs" permission in the manifest" https://developer.chrome.com/extensions/tabs

OfficialNoob:
Did a test it seems to work fine please review

content.js
```diff
@@ -1143,3 +1168,26 @@ function sendRequestToCustomServer(type, fullAddress, callback) {
   //submit this request
   xmlhttp.send();
 }
+
+function getYouTubeVideoID(url) { // Returns with video id else returns false
```
ajayyy: Don't add this back

popup.js
```diff
@@ -201,16 +201,18 @@ function runThePopup() {
...
   
-  
   chrome.tabs.query({
       active: true,
       currentWindow: true
-  }, loadTabData);
+  }, onTabs);
   
+  function onTabs(tabs) {
+	 chrome.storage.sync.get(['videoid'], function(result) {
```
ajayyy: change to `videoID`

background.js
```diff
@@ -1,27 +1,10 @@
-//when a new tab is highlighted
-chrome.tabs.onActivated.addListener(
-  function(activeInfo) {
-    chrome.tabs.get(activeInfo.tabId, function(tab) {
-      let id = getYouTubeVideoID(tab.url);
+//the id of this user, randomly generated once per install
+var userID = null;
```
ajayyy: Don't add this back

background.js
```diff
@@ -1,27 +1,10 @@
...
-      //if this even is a YouTube tab
-      if (id) {
-        videoIDChange(id, activeInfo.tabId);
-      }
-    })
-  }
-);
-
-//when a tab changes URLs
 chrome.tabs.onUpdated.addListener(function(tabId, changeInfo, tab) {
```
ajayyy: This can still be used without the permission?

content.js
```diff
@@ -179,11 +179,36 @@ document.onkeydown = function(e){
...
   //not a url change
-  if (sponsorVideoID == id){
-    return;
-  }
+  if (sponsorVideoID == id) return;
+  
+  chrome.storage.sync.set({videoid: id});
+
+  //warn them if they had unsubmitted times
+  if (previousVideoID != null) {
```
ajayyy: Content script can do notifications?

background.js
OfficialNoob: yes

content.js
OfficialNoob: Removed.

popup.js
OfficialNoob: Done

content.js
OfficialNoob: Not sure may need to create API I think the code works better per tab

OfficialNoob: Added alertPrevious message to background

OfficialNoob:
Made requested code changes
Note: I think it would be good sometime to have videoID for each tab in the storage API

ajayyy:
True, this will break if two tabs load at similar times. Can content scripts get their own tab id?

popup.js
```diff
@@ -201,16 +201,18 @@ function runThePopup() {
...
   chrome.tabs.query({
       active: true,
       currentWindow: true
-  }, loadTabData);
+  }, onTabs);
   
+  function onTabs(tabs) {
+	 chrome.storage.sync.get(['videoID'], function(result) {
+         	loadTabData(tabs, result.videoid);
+	 });
```
ajayyy: What is the point of this, why doesn't it send a message to the tab asking for the videoID?

popup.js
```diff
@@ -201,16 +201,18 @@ function runThePopup() {
...
+  function onTabs(tabs) {
+	 chrome.storage.sync.get(['videoID'], function(result) {
+         	loadTabData(tabs, result.videoid);
+	 }); 
+  }
   
-  function loadTabData(tabs) {
-    //set current videoID
-    currentVideoID = getYouTubeVideoID(tabs[0].url);
+  function loadTabData(tabs, currentVideoID) {
```
ajayyy: Is `tabs` still needed here?

OfficialNoob: Currently Yes as tabs[0].id is used for isInfoFound

popup.js
OfficialNoob: There might be problems when switching between multiple tabs

- Make Storage API contain the tabID
- Send a message for the active tab (Probably best option)

OfficialNoob:
Check now :/

popup.js
OfficialNoob: Went for "Send a message to active tab"

ajayyy:
So, the tabs message from the popup still works without the tabs permission?

background.js
```diff
@@ -1,27 +1,7 @@
-//when a new tab is highlighted
-chrome.tabs.onActivated.addListener(
```
ajayyy: Why was this removed?

ajayyy: This actually shouldn't be necessary anymore.

PR #92, opened 2019-08-08
Remove help/* from web_accessible_resources

bershanskiy:
Remove files in help/ directory from web_accessible_resources in manifest.json because the help page opens in context of the extension (no need to expose it to other contexts).


bershanskiy:
I yanked this out from #89 to make it easier to review.

ajayyy:
Do you think it would be a good idea to just load https://sponsor.ajay.app/help or keep it loaded locally?

bershanskiy:
I think loading the help page locally is a better move. I would also add a link to it somewhere in the UI so that the user can reference the help later.

ajayyy:
Merge first

PR #122, opened 2019-08-20
Added addButtons

OfficialNoob:
Status: [TESTING/REVIEW]
Currently uses wait util


content.js
```diff
@@ -529,53 +527,47 @@ function reskipSponsorTime(UUID) {
...
-}
-
 function removePlayerControlsButton() {
-    document.getElementById("startSponsorButton").style.display = "none";
+	if (!sponsorVideoID) return;
+    document.getElementById("changeStartSponsor").style.display = "none";
     document.getElementById("submitButton").style.display = "none";
 }
 
+function createButton(baseid, title, callback, imageName) {
```
ajayyy: `baseid` -> `baseID`

ajayyy:
This comment is for me:

Things to check:

- Removal of direct links doesn't break anything
- Adding buttons on channel trailers doesn't prevent them from being added to normal vids (I think this might be a problem)
- Maybe look into getting the controls in createButton instead of addButtons

content.js
OfficialNoob: Changed

content.js
```diff
@@ -211,20 +206,24 @@ function resetValues() {
     //reset sponsor times
     sponsorTimes = null;
     UUIDs = null;
-    sponsorVideoID = id;
     sponsorLookupRetries = 0;
 
     //empty the preview bar
-    previewBar.set([], [], 0);
+    previewBar = null;
```
ajayyy: Seems like a waste to keep creating new ones.

content.js
```diff
@@ -211,20 +206,24 @@ function resetValues() {
...
-    if (sponsorVideoID == id) return;
-
+    // ID has not changed return
+    if (sponsorVideoID === id) return
+	// Global ID
+	sponsorVideoID = id; 
+	resetValues();
+	
+	// ID is not valid
+	if (id === false) return;
```
ajayyy: Add a new line after this for spacing.

content.js
```diff
@@ -211,20 +206,24 @@ function resetValues() {
...
     sponsorDataFound = false;
 }
 
 function videoIDChange(id) {
-    //not a url change
-    if (sponsorVideoID == id) return;
-
+    // ID has not changed return
+    if (sponsorVideoID === id) return
+	// Global ID
```
ajayyy: Could you fix the tabbing here? 4 spaces

content.js
```diff
@@ -529,53 +530,53 @@ function reskipSponsorTime(UUID) {
...
-        setTimeout(addPlayerControlsButton, 100);
-        return;
-    }
-
-    referenceNode.prepend(startSponsorButton);
-}
-
 function removePlayerControlsButton() {
-    document.getElementById("startSponsorButton").style.display = "none";
+	if (!sponsorVideoID) return;
```
ajayyy: Can you fix the tabbing here and add some more new lines for spacing :)

content.js
```diff
@@ -529,53 +530,53 @@ function reskipSponsorTime(UUID) {
...
+    document.getElementById("changeStartSponsor").style.display = "none";
     document.getElementById("submitButton").style.display = "none";
 }
 
+function createButton(baseid, title, callback, imageName) {
+  if (document.getElementById(baseid+"Button") != null) return;
```
ajayyy: Can you fix the tabbing here and add some more new lines for spacing :)

Also, it seems you never changed `baseid` -> `baseID`.

content.js
```diff
@@ -529,53 +530,53 @@ function reskipSponsorTime(UUID) {
...
-}
-
 function removePlayerControlsButton() {
-    document.getElementById("startSponsorButton").style.display = "none";
+	if (!sponsorVideoID) return;
+    document.getElementById("changeStartSponsor").style.display = "none";
     document.getElementById("submitButton").style.display = "none";
 }
 
+function createButton(baseid, title, callback, imageName) {
```
ajayyy: What is title used for, seems like you are using `baseid` instead by accident.

content.js
```diff
@@ -529,53 +530,53 @@ function reskipSponsorTime(UUID) {
...
 function updateVisibilityOfPlayerControlsButton() {
     //not on a proper video yet
-    if (!getYouTubeVideoID(document.URL)) return;
-
-    addPlayerControlsButton();
-    addInfoButton();
-    addDeleteButton();
-    addSubmitButton();
+    if (!sponsorVideoID) return;
+
```
ajayyy: You should still try adding the buttons here as well.

content.js
```diff
@@ -620,13 +621,14 @@ function updateSponsorTimesSubmitting() {
     });
 }
 
+isSubmitLoaded = () => (document.getElementById("startSponsorImage") !== undefined);
```
ajayyy: Why is this called `isSubmitLoaded` but reference `startSponsorImage`. Also, don't you mean to check if it is equal to null, not undefined?

content.js
OfficialNoob: Title is just for the title however its using the i18n function

content.js
OfficialNoob: So is a previewBar on a page without a previewBar xD

content.js
OfficialNoob: sure

content.js
OfficialNoob: Ok

content.js
OfficialNoob: Seems like a bad idea running createButton even after its added

content.js
OfficialNoob: it is called that because the id startSponsorImage is added when createButton is run
if startSponsorImage is not undefined hopefully it exists

content.js
ajayyy: On the YouTube page, the video HTML object is still always there but just hidden.

content.js
OfficialNoob: yes

content.js
ajayyy: It is using baseid in the i18n function. I'm assuming that's a mistake?

content.js
ajayyy: They should have checks in the function

content.js
OfficialNoob: ok

content.js
ajayyy: it is never undefined, document.getElementByID returns null not undefined

ajayyy: But why not check for the submitButton. What does `isSubmitLoaded` mean?

content.js
OfficialNoob: document.querySelector("video") returned null tested in https://www.youtube.com/account

content.js
OfficialNoob: Yes it is

OfficialNoob: Changed

content.js
OfficialNoob: How will this change improve the extension?
The buttons should never get removed

content.js
OfficialNoob: It should return true once startSubmit button is added.
it needs to be changed to null not undefined

content.js
ajayyy: Yes, only if you start first on a video. But maybe not for pages like that. I have only tested logged out.

content.js
ajayyy: This is just in case the other function doesn't add them in time.

content.js
ajayyy: Then why doesn't it check if `submitButton` is not null?

_locales/en/messages.json
```diff
@@ -12,6 +12,15 @@
         "message": "Skip over sponsorship on YouTube videos. Report sponsors on videos you watch to save the time of others.",
         "description": "Description of the extension."
     },
+    "clearTimes": {
```
ajayyy: Can you add these to the bottom so it's obvious that translations are missing stuff

OfficialNoob: How would it be obvious that there missing stuff even if there at the bottom of the file

content.js
OfficialNoob: The change visibility function is used a lot I think its more likely to forget a check

_locales/en/messages.json
ajayyy: Since you can look at the bottom and see that the last ones are not the same.

content.js
OfficialNoob: @ajayyy it does now

_locales/en/messages.json
OfficialNoob: Moved however there may be problems with languages not being added at the same time

content.js
OfficialNoob: There are cases where this does not work

_locales/en/messages.json
ajayyy: Oh, will it not default back to English?

content.js
ajayyy: Shouldn't the preview bar still be reset. How is setting it to null going to clear it?

_locales/en/messages.json
OfficialNoob: I think it will however some translations may not be added at the same time

ajayyy: If a translation is missing one string, it defaults back to English right?

content.js
OfficialNoob: What do you mean? it gets removed

ajayyy: Setting it to null doesn't remove it from the page I don't think

OfficialNoob: It clears the object what did you want to happen

_locales/en/messages.json
OfficialNoob: yes

PR #123, opened 2019-08-20
Russian translation

bershanskiy:
Russian translation for all strings that are in `en/messsages.json` right now.

I tried to test it manually, but there are too many strings that are not translatable yet for that to make sense (I'd need to test it later anyway).


bershanskiy:
I'll also translate this into Ukrainian once more messages are moved to `en/messages.json`.

ajayyy:
So, do you think translations should not be added yet until it's all in messages.json. Like do you think partial translation will be too confusing?

bershanskiy:
@ajayyy Above all, partial translations are confusing for translators. Ideally, a translator would sit down once and do everything in one go to ensure consistency and to actually have enough time to pick the best words. If you translate in chunks (or multiple different people translate different messages), they are less likely to match each other. Also, if multiple people translate the thing, they usually give more automated translations similar to those returned by Google Translate.

## Glossary
You can translate "sponsor segment" in a few ways, but let's consider only two:
- "спонсорский **сегмент**" -- Google Translate suggests this one
_сегмент_ is borrowed or transliterated from English "segment", but in Russian it has a distinct geometric meaning [Толковый словарь Ожегова](https://gufo.me/dict/ozhegov/%D1%81%D0%B5%D0%B3%D0%BC%D0%B5%D0%BD%D1%82) defines it as (1) section of a line, (2) part of a circle or a sphere (3) one of multiple similar sections of an animal or organ (e.g., segment of a worm or a spinal cord). 
- or phrase "спонсорская **вставка**" ("sponsor **insert**") -- a native speaker would choose this one

This is not a deal-breaker (people will understand what you mean regardless), but the proper words give a more professional feel to the software. For some reason, if you translate sentences one by one, you are more likely to pick the transliterated word rather than translate the whole sentence.

## Consistency
Consider `hitGoBack` and `unskip`:
```
    "hitGoBack": {
        "message": "Hit unskip to get to where you came from."
    },
    "unskip": {
        "message": "Unskip"
    },
```
`unskip` is a button label and `hitGoBack` references it by name. You can translate "Unskip" in a few _slightly_ different ways, but a good translation would be consistent.

ajayyy:
Yep, I agree with you. I was just doing a French translation and having similar issues. So, do you think I should hold off all translations until all parts are in messages.json?

bershanskiy:
> So, do you think I should hold off all translations until all parts are in messages.json?

If I'm not mistaken, you plan to redesign the UI significantly, right? If so, the messages are subject to change and it probably makes no sense to translate everything yet.

bershanskiy:
@ajayyy Did you see my comment above about `Mins` and `Secs`?

ajayyy:
@bershanskiy exactly, I was planning on not adding translation support to the current UI and only the new one.

ajayyy:
@bershanskiy I don't see a message about `Mins` and `Secs`

ajayyy:
Can you add this one missing translation?

https://github.com/ajayyy/SponsorBlock/blob/experimental-ajay/_locales/en/messages.json#L127

bershanskiy:
@ajayyy 
> Can you add this one missing translation?

Done.

ajayyy:
Firefox titles have a 50 char limit. Is there a shorter way the title can be written and still be understandable?

bershanskiy:
@ajayyy Are you talking about `fullName`?
If so, you could write "SponsorBlock - Проматывайте спонсоров на YouTube". It barely fits the limit but is comprehensible and non-ambiguous.

ajayyy:
Yes, I am, thanks!

PR #130, opened 2019-08-22
Better french translation

doubouil:
Native speaker here, the machine based translation was not great so here's a first draft.
I choose to use "_message commercial_" (commercial message) and not "_publicité_" (advertisement) since ads feels to me to be more a legal status than a definite thing. "Endossement" is not a valid french word in this context, and endorsements would be translated to "_sponsors_" or "_mécénat_" in fr_FR, but again this are more legal category than pure identifying terms and fr_CA strongly dislikes frenglish.
But I left it in the extension name since this is a good keyword for people looking for adblocking-related extensions.

### DIFF
--- a/_locales/fr/messages.json
+++ b/_locales/fr/messages.json
@@ -4,7 +4,7 @@
         "description": "Name of the extension."
     },
     "fullName": {
-        "message": "SponsorBlock pour YouTube - Enlève les messages commerciaux et publicités",
+        "message": "SponsorBlock pour YouTube - Enlève les messages commerciaux et publicités intégrées",
         "description": "Name of the extension."
     },
 
@@ -43,7 +43,7 @@
         "message": "segments de message commercial"
     },
     "noticeTitle": {
-        "message": "Passer ce message"
+        "message": "Message passé"
     },
     "reportButtonTitle": {
         "message": "Incorrect"
@@ -64,13 +64,13 @@
         "message": "Secondes"
     },
     "Hide": {
-        "message": "Toujours cacher"
+        "message": "Ne plus montrer"
     },
     "hitGoBack": {
-        "message": "Revenir au début du segment"
+        "message": "Cliquez sur revenir en arrière pour revenir avant le saut du segment"
     },
     "unskip": {
-        "message": "Ne pas sauter"
+        "message": "Revenir en arrière"
     },
     "reskip": {
         "message": "Sauter"
--- a/_locales/fr/messages.json
+++ b/_locales/fr/messages.json
@@ -4,12 +4,12 @@
         "description": "Name of the extension."
     },
     "fullName": {
-        "message": "SponsorBlock pour YouTube - Enlève les endossements",
+        "message": "SponsorBlock pour YouTube - Enlève les messages commerciaux et publicités",
         "description": "Name of the extension."
     },
 
     "Description": {
-        "message": "Enlève les endossements dans les vidéos YouTube. Soummettre les endossements dans les vidéos que vous regardez pour aidez les autres.",
+        "message": "Passe automatiquement les messages commerciaux intégrés dans les vidéos YouTube. Soumettez les segments commerciaux dans les vidéos que vous regardez pour aidez les autres.",
         "description": "Description of the extension."
     },
     "helpPage": {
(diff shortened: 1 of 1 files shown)


_locales/fr/messages.json
```diff
@@ -61,69 +61,69 @@
...
-        "message": "Ne Montre Jaimais"
+        "message": "Toujours cacher"
     },
     "hitGoBack": {
-        "message": "Clique retourne pour si vous avez manqué parti."
+        "message": "Revenir au début du segment"
     },
     "unskip": {
-        "message": "Retourne"
+        "message": "Ne pas sauter"
```
ajayyy: I don't know about this one, this is for the action of going back to the beginning after the skip.

_locales/fr/messages.json
```diff
@@ -61,69 +61,69 @@
...
-        "message": "Seconds"
+        "message": "Secondes"
     },
     "Hide": {
-        "message": "Ne Montre Jaimais"
+        "message": "Toujours cacher"
     },
     "hitGoBack": {
-        "message": "Clique retourne pour si vous avez manqué parti."
+        "message": "Revenir au début du segment"
```
ajayyy: This message should be telling you to click the "unskip" button to return. This is the message telling you how to get back.

_locales/fr/messages.json
```diff
@@ -19,40 +19,40 @@
...
-        "message": "section d'endossement"
+        "message": "segment de message commercial"
     },
     "Segments": {
-        "message": "section d'endossements"
+        "message": "segments de message commercial"
     },
     "noticeTitle": {
-        "message": "Endossement Passer"
+        "message": "Passer ce message"
```
ajayyy: This should be past-tense, as the message was just skipped.

_locales/fr/messages.json
```diff
@@ -4,12 +4,12 @@
...
     },
     "fullName": {
-        "message": "SponsorBlock pour YouTube - Enlève les endossements",
+        "message": "SponsorBlock pour YouTube - Enlève les messages commerciaux et publicités",
         "description": "Name of the extension."
     },
 
     "Description": {
-        "message": "Enlève les endossements dans les vidéos YouTube. Soummettre les endossements dans les vidéos que vous regardez pour aidez les autres.",
+        "message": "Passe automatiquement les messages commerciaux intégrés dans les vidéos YouTube. Soumettez les segments commerciaux dans les vidéos que vous regardez pour aidez les autres.",
```
ajayyy: `Passe` was the word I was looking for, I went with `Enlève` since I forgot, thanks!

_locales/fr/messages.json
```diff
@@ -4,12 +4,12 @@
         "description": "Name of the extension."
     },
     "fullName": {
-        "message": "SponsorBlock pour YouTube - Enlève les endossements",
+        "message": "SponsorBlock pour YouTube - Enlève les messages commerciaux et publicités",
```
ajayyy: What do you think about `messages commerciaux et publicités intégrés`?

_locales/fr/messages.json
```diff
@@ -61,69 +61,69 @@
         "message": "Minutes"
     },
     "Secs": {
-        "message": "Seconds"
+        "message": "Secondes"
     },
     "Hide": {
-        "message": "Ne Montre Jaimais"
+        "message": "Toujours cacher"
```
ajayyy: Is this understood as "Don't show again"?

_locales/fr/messages.json
doubouil: In the sense of "Oops I wanted to see that" this could be "Revenir en arrière" (go back) but it feels a bit long. Since the skip/jump is always forward by default, "un-jump" is my understanding but "Sauter en arrière" is more like "jump backward" than "jump back". So I chose a short one with "Ne pas sauter" which indicates more the "I didn't/don't want that" than "do this now".

_locales/fr/messages.json
doubouil: A better translation would be "Cliquez ici pour revenir avant le saut du segment" (click here to go back before sponsor skip). I haven't seen this string in the UI so excuse my confusion.

_locales/fr/messages.json
doubouil: "Message passé" ("Sponsor skipped")

_locales/fr/messages.json
doubouil: "Enleve" is hard remove, "passe" is skip :)

_locales/fr/messages.json
doubouil: "messages commerciaux et publicités intégré**e**s" since ads are "female"

_locales/fr/messages.json
doubouil: "Ne plus montrer" would be "Dont show again" and seems better indeed.

_locales/fr/messages.json
ajayyy: Is `retourne` not a good word for it?

_locales/fr/messages.json
ajayyy: Here is where it appears. It is the message that appears after hitting the report button. The long one in the middle.

_locales/fr/messages.json
ajayyy: Yep, I like "Ne plus montrer" a lot better.

_locales/fr/messages.json
doubouil: "Retourne" is not a noun, nor a verb that can be displayed as is, this would be "Retournez" (go back) but without anything else we don't know where we go back. "Retourne" is also "flipping it over" so "Revenir" is better, more temporal. Back is "en arrière" so "Revenir en arrière" is "Go back". Again this indicates what the button does, but not the context that the result of the previous action might not have been what we expected, but it's a very slim line and I think people would understand both.

_locales/fr/messages.json
doubouil: Ok so if the button is "Revenir en arrière" this text would be "Cliquez sur Revenir en arrière pour revenir avant le saut du segment"

PR #134, opened 2019-08-24
Content.js messageListener switch

OfficialNoob:


ajayyy:
Why `return` vs `break`?

OfficialNoob:
Just because its the only thing in the function using return allows messageListener to return stuff not sure if its faster or not

PR #135, opened 2019-08-24
Popup i18n support

OfficialNoob:


ajayyy:
You might want to look at https://stackoverflow.com/a/39810769/1985387 which leaves default text there at the beginning. Though, if the JS is fast enough, it is not needed.

OfficialNoob:
@ajayyy this PR seems to be done

PR #153, opened 2019-10-15
Some Firefox Optimizations

owendaprile:
This stops Firefox from throwing an error every time user changes tabs if the addon is installed at all. It also moves the browser specific settings into the main manifest file which only causes Chrome to throw a warning when loading the unpacked extension and with no changes when installing the .crx.

### DIFF
--- a/background.js
+++ b/background.js
@@ -1,7 +1,7 @@
-chrome.tabs.onUpdated.addListener(function(tabId, changeInfo, tab) {
+chrome.tabs.onUpdated.addListener(function(tabId) {
 	chrome.tabs.sendMessage(tabId, {
         message: 'update',
-	});
+	}, () => void chrome.runtime.lastError ); // Suppress error on Firefox
 });
 
 chrome.runtime.onMessage.addListener(function (request, sender, callback) {
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 config.js
 ignored
+.idea/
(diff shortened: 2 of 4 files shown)


background.js
```diff
@@ -1,7 +1,7 @@
-chrome.tabs.onUpdated.addListener(function(tabId, changeInfo, tab) {
+chrome.tabs.onUpdated.addListener(function(tabId) {
 	chrome.tabs.sendMessage(tabId, {
         message: 'update',
-	});
+	}, () => void chrome.runtime.lastError ); // Suppress error on Firefox
```
ajayyy: What does `void` do?

owendaprile: AFAIK, Firefox only prints the error to the console when it doesn’t find any code that handles the error so this is just a callback that sends the error to [void](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void) which evaluates it and then returns undefined.

ajayyy: Ah, that is very cool 👍

Mathnerd314: I think it would be cleaner to use [extraParameters](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/Tabs/onUpdated#addListener_syntax) to filter down to just the tabs where the content script is injected. I don't know how Chrome handles the extra parameter.

ajayyy: Good idea, I made an issue: https://github.com/ajayyy/SponsorBlock/issues/155

ajayyy:
The thing with the manifest is that Firefox does not need it when you publish to the store, only when testing. Since I test on chrome, I would rather keep it out and only leave it there for when Firefox tests are needed.

PR #163, opened 2019-10-31
Keybind edit, show notice again, ignore rate limits

ajayyy:


content.js
```diff
@@ -80,13 +80,22 @@ chrome.storage.sync.get(["trackViewCount"], function(result) {
 
 //if the notice should not be shown
 //happens when the user click's the "Don't show notice again" button
+//option renamed when new notice was made
 var dontShowNotice = false;
-chrome.storage.sync.get(["dontShowNoticeAgain"], function(result) {
+chrome.storage.sync.get(["dontShowNotice"], function(result) {
     let dontShowNoticeAgain = result.dontShowNoticeAgain;
```
Mathnerd314: This is wrong. Storage returns `{dontShowNotice: true}` so `result.dontShowNoticeAgain` is undefined and it shows the notice every time.

ajayyy: Oops, I fixed this in popup.js but forgot here.

PR #164, opened 2019-10-31
Improve French translation

Otiel:
A few fixes, improvements and missing translations to the french localization strings.

### DIFF
--- a/_locales/fr/messages.json
+++ b/_locales/fr/messages.json
@@ -19,7 +19,7 @@
         "message": "Soumission invalide"
     },
     "429": {
-        "message": "Vous cherchez à envoyer beaucoup de segments, il y en a vraiment autant ?"
+        "message": "Vous cherchez à envoyer beaucoup de segments, y en a-t-il vraiment autant ?"
     },
     "409": {
         "message": "Déja soumis"
@@ -28,7 +28,7 @@
         "message": "Le serveur ne fonctionne pas. Contactez le développeur."
     },
     "channelWhitelisted": {
-        "message": "Cette chaine est sur la liste blanche !"
+        "message": "Cette chaîne est sur la liste blanche !"
     },
     "Sponsor": {
         "message": "message commercial"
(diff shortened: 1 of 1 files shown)


ajayyy:
Thank you! I will review it in a bit.

_locales/fr/messages.json
```diff
@@ -109,21 +109,158 @@
...
+    },
+    "noticeUpdate2": {
+        "message": "Si elle ne vous plaît pas, cliquez sur le bouton \"Ne plus montrer\"",
+        "description": "The second line of the message displayed after the notice was upgraded."
+    },
+    "setStartSponsorShortcut": {
+        "message": "Changer le raccourci pour enregistrer un segment"
+    },
+    "setSubmitKeybind": {
+        "message": "Changer le raccourci pour soumettre les segments"
```
ajayyy: Is `raccourci` the best word to use here? Maybe `clé`?

_locales/fr/messages.json
```diff
@@ -109,21 +109,158 @@
...
+    },
+    "keybindDescription": {
+        "message": "Appuyez sur une touche"
```
ajayyy: Would `clé` be better here?

_locales/fr/messages.json
Otiel: No `key` is a false friend in French.

* Keyboard = clavier
* Key (from keyboard) = touche
* _Key (to open a door) = clé_
* Keyboard shortcut(s) = touche(s) de raccourci / raccourci(s) clavier

And like in English you often hear "shortcut" only (not keyboard shortcut), we usually say only "raccourci" when the context is enough to understand that we're talking about the keyboard.

_locales/fr/messages.json
Otiel: See [my other comment](https://github.com/ajayyy/SponsorBlock/pull/164#discussion_r342191375): `raccourci` is the right term here.

_locales/fr/messages.json
ajayyy: Okay, thanks for the explanation. I had not heard that term before.

PR #169, opened 2019-11-14
Decreased requests to help with server load issues

ajayyy:


Chess13234:
Can the extension bundle the db for install to reduce request load?

ajayyy:
@Chess13234 this is something I want to do, but it will hurt the real time part of the extension. I am thinking of ignoring the local db if the video is recently uploaded

Issue #61, opened 2019-08-02
Support invidious


ajayyy:

Z3ROGRAV1TYx:
Bumping this, would absolutely love to see [Invidious](https://github.com/omarroth/invidious) support please!

tuxayo:
I suppose that many people knowing SponsorBlock and Invidious would like to use booth. So they are cannibalizing each other's userbase as one must choose one.

Z3ROGRAV1TYx:
@tuxayo wdym?

ajayyy:
Finished

tuxayo:
I just tested, this amazing!!!!!! :D
Thank you! thank you! thank you!

tuxayo:
@Z3ROGRAV1TYx 
> @tuxayo wdym?

For example, I couldn't use SponsorBlock because I used Invidious. And I guess a lot of people using SponsorBlock couldn't use Invidious even though they wanted. So it was splitting the user base. But not anymore!!! :D

PR #181, opened 2019-12-05
Initial Invidious support

afrmtbl:
Adds initial support for Invidious

Few things to note:
~1. It only works on `invidio.us` and `invidiou.sh` since it's not really possible to detect whether a site is an Invidious instance using just the manifest (as far as I know anyways).~
2. It does not add the buttons to the controls area because there isn't really a good place for them.
3. I only tested on Chromium 78

Addresses #61


ajayyy:
Oooh, thank you. I really like that for loop, it will be useful if there are anyone sites integrate.

I wanted to setup some [optional permissions](https://stackoverflow.com/a/26677279/1985387) for this to prevent the extension from being disabled on update. Would you be interested in adding that, if not I will :)

Thanks again for this :)

afrmtbl:
I think it'd be better if you implemented it. I'm not too familiar with the codebase yet or `optional_permissions` in general :p

TheFrenchGhosty:
>     1. It only works on `invidio.us` and `invidiou.sh` since it's not really possible to detect whether a site is an Invidious instance using just the manifest (as far as I know anyways).

You can use https://instances.invidio.us/ the same way [Invidition](https://codeberg.org/Booteille/Invidition) uses it.

Perflyst:
If this is not that easily possible maybe an option to set a custom instance would be a workaround for the moment too

ajayyy:
Having it automatic would require more permissions (read ALL sites visited). So, it is probably best not to do that

ajayyy:
Waiting on custom options page

ajayyy:
The notice has a slight visual issue:

Invidious:
image

Normal:
image

ajayyy:
#216

ajayyy:
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/contentScripts/register

ajayyy:
This **does not** work in incognito due to [a chrome bug](https://bugs.chromium.org/p/chromium/issues/detail?id=825507)

Perflyst:
Since you decided to hard-code the invidious instances, I would like to ask you to add `invidious.snopyta.org` too.

Caboose700:
Will there be any way to manually specify a invidious instance? This would be useful for those of us who host their own private invidious that aren't meant to be accessed publicly.

ajayyy:
Yes

ajayyy:
It seems there are somehow even bigger css issues with the notice now.

PR #190, opened 2019-12-09
 Add a gecko ID to the extension

TheFrenchGhosty:
Some browsers (like GNU Icecat) will mark an add-on as corrupt if the manifest.json doesn't contain a gecko ID.

This PR fixes that.

For more details see: https://github.com/ysard/cookie-quick-manager/issues/75


TheFrenchGhosty:
I just noticed that a gecko ID is included in https://github.com/ajayyy/SponsorBlock/blob/master/firefox_manifest-extra.json but sadly it isn't working that way and it really needs to be in the manifest.json

ajayyy:
Doesn't the addons store add it automatically. It doesn't work in chrome if added (that's why it is not there)

TheFrenchGhosty:
Sadly no it needs to be added manually into the file... I guess the solution is to do 2 different manifest.json, one for each browser.

ajayyy:
Are you downloading it from the addons store or from the repo?

TheFrenchGhosty:
I tried both, the problem is here for both.

See this gif for the store:

sponsorblock

ajayyy:
This applies to Firefox versions less than 48: https://stackoverflow.com/a/56823677/1985387

ajayyy:
Latest extension now has it

TheFrenchGhosty:
Amazing! It can now be installed, thanks!

PR #208, opened 2019-12-25
Add `prettier` for code formatting

jplsek:
There were comments in some PRs/Issues about code style.

I choose to add a prettifier to the repository. Later, this can be added to a CI to check that PRs have good code style.

Another alternative is to use `eslint`, but `prettier` is very simple to use, and `eslint` could be added later to find bugs.

The current settings chosen are using the current "trend" for js/html code style. The big thing compared to this repo is: 2 spaces instead of tabs and no semicolons. Because of this, I choose not to run the formatter because every file would change. I'd rather let @ajayyy run the formatter and create a commit on top of this PR.

Or... We can just change the options to use tabs and semicolons. @ajayyy I'd like your opinion on this.


ajayyy:
Awesome, I'll check it out when I'm on my PC. I changed the style to my style (due to me being used to Java).

ajayyy:
Any `yarn` commands can be substituted for `npm`, right?

jplsek:
For the most part, yes. I recommend using `yarn` because it is much faster and smarter (flatter and smaller `node_modules` dir) when installing dependencies. This is mostly noticeable for larger projects, but this still helps for CI and general development.

ajayyy:
Is there a way to make this not happen:

```
chrome.tabs.query(
    {
        active: true,
        currentWindow: true,
    },
    tabs => {
        chrome.tabs.sendMessage(tabs[0].id, {
            message: "sponsorDataChanged",
        });
    }
);
```
Originally:
```
chrome.tabs.query({
        active: true,
        currentWindow: true,
}, tabs => {
        chrome.tabs.sendMessage(tabs[0].id, {
            message: "sponsorDataChanged",
        });
});
```

jplsek:
Hmmm, `prettier` is very opinionated on purpose, so there aren't many configuration options and the configurations I tried couldn't get those results. For js, `eslint` can be configured to enforce practically any style (there's hundreds of rules). Maybe I can look into creating a configuration similar to your style that uses eslint for js? Prettier can be used for everything else.

jplsek:
Okay, added an eslint configuration. Since eslint is typically used to find bugs, it will spit out errors related to possible issues, which have to fixed manually (double vs triple equals, global variables that were not found, etc). The rules can be configured further to ignore them though.

ajayyy:
Awesome! Do they both get used together or should prettier be removed?

jplsek:
Since eslint is only for js, prettier is used for the css and html. So both are used together.

OfficialNoob:
For release I think it should be minified. (Website or extension)
However for development it would be good having standardized formatting.

ajayyy:
Yes, compilation would help solve https://github.com/ajayyy/SponsorBlock/pull/190

jplsek:
I looked into automating artifacts with the help of web-ext and github actions. See https://github.com/jplsek/SponsorBlock/commit/559a2573215cf4f3623d5597d3ec80514f088933 and https://github.com/jplsek/SponsorBlock/runs/367202916. It's not minified, but it's a step.

I was originally going to wait until this got merged because of adding npm/yarn, but I'll look into creating a PR soon instead of waiting. Then just rebase which ever gets merged first.

jplsek:
Rebased with master and cleaned some commits.

ajayyy:
Can you avoid force pushing in the future?

jplsek:
ok v.v (Feel free to message me on discord to help deal with / work around it).

ajayyy:
Any idea how to fix:

```
~/utils.js
  65:35  error  Parsing error: Assigning to rvalue
```

jplsek:
`obj.setAttribute(attribute) = localizedMessage` I think the linter is correct. That looks like a bug to me?

As for the newlines added, I'm not sure besides adding `/* prettier-ignore */` above the line. (Also the font-family should not be needed since that's already inherited from the body. It seems every instance font-family is used like that and can be removed).

ajayyy:
@jplsek oh my, for some reason I didn't notice that it gave a line number for that error. Sorry about that. That is indeed an error. Thanks!

phiresky:
I'd recommend you add a prettier pre-commit hook: https://prettier.io/docs/en/precommit.html

The hook is auto installed when you first run `npm install`.

That way whenever you create a commit the code is auto-formatted. This prevents the code from deviating over time because people forget to run the format command. I'm using it in many repos with great success, no more formatting inconsistencies.

jplsek:
That's a good idea. My original plans was that if this PR got merged, I was gonna update the CI to run this check (which I'd still do).

ajayyy:
I think I'm going to try to convert the project to type script before merging this. I'll tell you if I decide against that idea.

jplsek:
We could remove the eslint formatter from this PR and revisit it later, then we at least have formatting for html and css.

phiresky:
> I think I'm going to try to convert the project to TypeScript

Awesome! Do you need help? I have lots of experience with TypeScript, though I've only written one (incomplete) browser extension myself with it.

I'd recommend you use [webextension-polyfill](https://github.com/mozilla/webextension-polyfill) with [@types/firefox-webext-browser](https://www.npmjs.com/package/@types/firefox-webext-browser), packaged with webpack.

this way you can use all the `chrome.*` APIs with promises/await instead of callbacks.

Or maybe [webextension-polyfill-ts](https://github.com/Lusito/webextension-polyfill-ts) looks good, not tried that before.

ajayyy:
I don't want the rewrite to be too big for now. However, it's a good idea
to add that so that it can be used in the future.

ajayyy:
Sorry for never merging this. I want to just stick with no auto-formatting for now.

PR #210, opened 2019-12-28
Confirm on video that is not public [TESTING]

OfficialNoob:
# Summary
- Added confirmPrivacy message to en
- Changed videoIDChange to async so that it can now use await
- Added Privacy utils in the content script
- return videoIDChange if user clicks cancel (Does not apply to /embed/)

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -320,7 +320,7 @@
     },
     "statusReminder": {
         "message": "Check status.sponsor.ajay.app for server status."
-    }
+    },
     "confirmPrivacy": {
         "message": "Check status.sponsor.ajay.app for server status."
     }
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -321,4 +321,7 @@
     "statusReminder": {
         "message": "Check status.sponsor.ajay.app for server status."
     }
+    "confirmPrivacy": {
+        "message": "Check status.sponsor.ajay.app for server status."
+    }
 }
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -322,6 +322,6 @@
         "message": "Check status.sponsor.ajay.app for server status."
     },
     "confirmPrivacy": {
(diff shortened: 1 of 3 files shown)


OfficialNoob:
isPublic does not work for embedded videos

ajayyy:
Does this appear right away or is it lazy loaded? If so, is a promise needed?

ajayyy:
Maybe add a "always check for sponsors" button to the dialog. Is this possible with vanilla confirmation menus?

ajayyy:
Does this create any noticeable extra delay when pulling sponsors? Also, does this work for the old YouTube theme?

OfficialNoob:
@ajayyy I have tested it for the normal theme :D

OfficialNoob:
Please review the code changes. the UI can be added later just want to warn users currently

ajayyy:
Confirmed to slow sponsor checking. This will have to be optional I guess. Then it can be removed in favour of https://github.com/ajayyy/SponsorBlockServer/issues/25

OfficialNoob:
@ajayyy It is slower now but I dont know how to make it faster without using a hashed substring or a synced local database

ajayyy:
@OfficialNoob yea, there is no way. Hashed substring is the best option.

OfficialNoob:
I think this is done until the new api endpoint is available

ajayyy:
Does this video break it: https://www.youtube.com/watch?v=tO8aJ-TUtJY&list=PL96C35uN7xGJu6skU4TBYrIWxggkZBrF5&index=1 ("Learning playlist")

OfficialNoob:
@ajayyy it did I have now added an exception for that type.
Im not sure if I should use a blacklist or a whitelist for detection

OfficialNoob:
Switched to only alerting on "Private" and "Unlisted"

ajayyy:
It is really only unlisted that is concerning

OfficialNoob:
Will recreate PR due to code changes

PR #215, opened 2019-12-30
Add web-ext and ci artifact generation

jplsek:
This adds a CI using GitHub Actions to generate build artifacts of the extension.

This generates artifacts for both Chrome and Firefox.

This uses `web-ext` to generate the artifacts. web-ext is also handy for running the extension under a clean Firefox profile. However, if the project doesn't want to add web-ext, and instead do something more custom, like running a script that just zips everything relevant, then that is fine, I can look into that.

See https://github.com/jplsek/SponsorBlock/runs/367214841 (top right is the artifacts button)

I'm not exactly sure what other files should not be included in the zip. Let me know (or make a commit). (see the `-i` argument in the package.json build script)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,38 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+
+  build:
+    name: Create artifacts
+    runs-on: ubuntu-latest
+
+    steps:
+      # Initialization
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v1
+      - run: npm install
+      - name: Copy configuration
+        run: cp config.js.example config.js
+
+      # Create Chrome artifacts
+      - name: Create Chrome artifacts
+        run: npm run build
+      - uses: actions/upload-artifact@v1
+        with:
+          name: Chrome Extension
+          path: web-ext-artifacts
+
+      # Create Firefox artifacts
+      - name: Move manifest
+        run: mv manifest.json manifest.json.original
+      - name: Combine manifest for Firefox
+        run: jq -s '.[0] * .[1]' manifest.json.original firefox_manifest-extra.json > manifest.json
+      - name: Create Firefox artifacts
+        run: npm run build
+      - uses: actions/upload-artifact@v1
+        with:
+          name: Firefox Extension
+          path: web-ext-artifacts
+
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ There are also other build scripts available. Install `npm`, then run `npm insta
 
 ## Developing with a clean profile
 
-Run `npm run dev` to run the extension using a clean Firefox profile with hot reloading.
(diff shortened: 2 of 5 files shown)


ajayyy:
Oh wow, this is amazing! Thanks for doing all of this, I am learning a lot from your PRs.

What do you mean by "a clean Firefox profile"?

jplsek:
Using the `npm run dev` script, `web-ext` by default will launch a temporary Firefox profile with nothing set (like running Firefox for the first time after an installation) except with the extension already installed. It will auto reload the extension for you (background.js) whenever a file is saved during development. There is also an option to start the console on start up (`--bc`). You can use an existing Firefox profile by also providing the `-p` flag.

README.md
```diff
@@ -54,7 +54,7 @@ There are also other build scripts available. Install `npm`, then run `npm insta
 
 ## Developing with a clean profile
 
-Run `npm run dev` to run the extension using a clean Firefox profile with hot reloading.
+Run `npm run dev` to run the extension using a clean browser profile with hot reloading [(by default Firefox)](https://hacks.mozilla.org/2019/10/developing-cross-browser-extensions-with-web-ext-3-2-0/). This uses [`web-ext run`](https://extensionworkshop.com/documentation/develop/web-ext-command-reference/#commands).
```
jplsek: Huh. That's new. Didn't know that it supported Chrome now. Sweet.

ajayyy:
For future reference: https://github.com/marketplace/actions/chrome-addon-action

jplsek:
And https://github.com/marketplace/actions/firefox-addon-action :). Since those deal with auth keys, I'll let you handle that ;). You'll just have to create a new workflow with an ["on: release"](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/events-that-trigger-workflows#release-event-release) type event.

ajayyy:
NOTE: `npm` uses `--` to use options unlike `yarn`. Ex. `npm run dev -- -bc -p $PROFILE --start-url $YOUTUBE_VIDEO`

ajayyy:
https://github.com/marketplace/actions/upload-a-release-asset

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
...
+      - uses: actions/upload-artifact@v1
+        with:
+          name: Chrome Extension
+          path: web-ext-artifacts
+
+      # Create Firefox artifacts
+      - name: Move manifest
+        run: mv manifest.json manifest.json.original
+      - name: Combine manifest for Firefox
+        run: jq -s '.[0] * .[1]' manifest.json.original firefox_manifest-extra.json > manifest.json
```
ajayyy: Is there a reason `*` is used instead of `+`?

jplsek: Not really, no. I didn't know about the `+` operator at the time.

If the files have the same keys where the vales are objects, then using `*` will merge all keys in both objects into one object. Using `+` will only merge the right hand side.

For example:
```json
{"k": {"a": 0}} + {"k": {"b": 1}} -> {"k":{"b":1}}
{"k": {"a": 0}} * {"k": {"b": 1}} -> {"k":{"a":0,"b":1}}
```

In this situation, using `+` would have been fine.

PR #216, opened 2019-12-30
Options Page

ajayyy:
Resolves https://github.com/ajayyy/SponsorBlock/issues/200

This page will be much easier and make it super easy to add new options.

image


help/index_en.html
```diff
@@ -13,7 +13,7 @@
 
   <center>
 
-    <p class="createdBy">Created By <a href="https://ajay.app">Ajay Ramachandran</a></p>
+    <p class="createdBy">Created By <a href="https://ajay.app">Ajay Ramachandran</a> <img src="https://ajay.app/newprofilepic.jpg" height="30" class="profilepiccircle"/></p>
```
jplsek: Since the permissions aren't set for just ajay.app (and to avoid adding extra unnecessary permissions), you may want the image to be in the repo/extension.

ajayyy: Hmm, I though I changed that. I did add it to the icons folder but I guess I forgot to change the link.

options/options.html
```diff
@@ -0,0 +1,149 @@
...
+</head>
+
+<body class="sponsorBlockPageBody">
+
+	<div id="title">
+		<img src="../icons/LogoSponsorBlocker256px.png" height="80" class="profilepic"/>
+		SponsorBlock
+	</div>
+
+  	<center>
```
jplsek: Center is technically not supported. You should use css for this.

help/index_en.html
ajayyy: https://github.com/ajayyy/SponsorBlock/issues/218

PR #219, opened 2019-12-31
Config proxy [WIP]

OfficialNoob:
# Global config
Having a global SB.config that can be used to store, get and set variables easily.
Initiation:  `<script src="SB.js"></script>` in `<head>` already in content scripts :D
Example: `SB.config.userID = "blah";`
Debug: `SB.localconfig` (this should store all settings may be useful to enumerate in console)
Default values: `SB.defaults` from `SB.js`
Reset to default: `resetConfig()`


OfficialNoob:
The code would be better in utils but its only a concept

popup.js
```diff
@@ -1,8 +1,41 @@
+SB = {};
+
+function configProxy() {
+    chrome.storage.onChanged.addListener((changes, namespace) => {
+        for (key in changes) {
```
ajayyy: `for (key of changes)`

OfficialNoob: I took that from https://developer.chrome.com/extensions/storage

OfficialNoob: Should it be changed?

ajayyy: You are right. It seems `in` goes through a JSON object and `of` goes through an array.

OfficialNoob:
I might try to do the Initiation in utils.js to safe time

OfficialNoob:
Trying to get utils.js to load the sync code sooner

content.js
```diff
@@ -72,51 +72,42 @@ var popupInitialised = false;
...
+	
 //load the legacy option to hide the notice
 var dontShowNoticeOld = false;
-chrome.storage.sync.get(["dontShowNoticeAgain"], function(result) {
-    let dontShowNoticeAgain = result.dontShowNoticeAgain;
-    if (dontShowNoticeAgain != undefined) {
-        dontShowNoticeOld = dontShowNoticeAgain;
-    }
-});
+let dontShowNoticeAgain = SB.config.dontShowNoticeAgain;
```
ajayyy: This will cause an error from multiple definitions.

OfficialNoob: Outdated

OfficialNoob:
Added run_at for https://github.com/ajayyy/SponsorBlock/issues/221

OfficialNoob:
@ajayyy despite the amount of code changes it seems to still work.
testing might take a while

OfficialNoob:
SB.config.sponsorTimes.set() VS SB.config["sponsortime"+key]

OfficialNoob:
Close was a mistake

SB.js
```diff
@@ -0,0 +1,67 @@
...
+            })
+        },
+        get: function(obj, prop) {
+            return SB.localconfig[prop]
+        }
+    };
+    return new Proxy({}, handler);
+}
+
+fetchConfig = _ => new Promise(function(resolve, reject) {
```
ajayyy: Can you hange this to `(_) =>` or just `() =>` for clarity?

SB.js
```diff
@@ -0,0 +1,67 @@
...
+    chrome.storage.sync.get(null, function(items) {
+        SB.localconfig = items;  // Data is ready
+        resolve();
+    });
+});
+
+async function config() {
+    await fetchConfig();
+	addDefaults();
```
ajayyy: Can you use spaces here?

content.js
```diff
@@ -632,42 +540,30 @@ function skipToTime(v, index, sponsorTimes, openNotice) {
 
     if (openNotice) {
         //send out the message saying that a sponsor message was skipped
-        if (!dontShowNotice) {
-            let skipNotice = new SkipNotice(this, currentUUID, disableAutoSkip);
+        if (!SB.config.dontShowNotice) {
+            let skipNotice = new SkipNotice(this, currentUUID, SB.config.disableAutoSkip);
 
-            if (dontShowNoticeOld) {
+            if (!SB.config.dontShowNoticeOld) {
```
ajayyy: This old notice code can be removed now.

SB.js
ajayyy: I'll fix this after, don't worry about it.

SB.js
OfficialNoob: Yes

content.js
OfficialNoob: Removed

OfficialNoob:
> SB.config

Is this needed as It would effect performance?

OfficialNoob:
@ajayyy this could be used for migration
```
function migrate() {
    for (key in SB.localconfig) {
        if (key.startWith("sponsortime")) {
            SB.config.set(key.substr(12), SB.config[key]);
            delete SB.config[key];
        }
    }
}
```

ajayyy:
Either method can be used. They both should work.

OfficialNoob:
@ajayyy Done

OfficialNoob:
A bit to soon last time

ajayyy:
Editing a submitted sponsor time doesn't seem to work.

ajayyy:
Current state: editing sponsor times still doesn't work since chrome does **not** store custom classes properly. How did you get the map working?

OfficialNoob:
Added encoding for maps however .set does not seem to cause an event

SB.js
```diff
@@ -0,0 +1,97 @@
...
+	return JSON.stringify(data);
+}
+
+function strParser(data) {
+	try {
+        return new Map(JSON.parse(data));
+    } catch(e) {
+        return data
+    }
+}
```
ajayyy: Change try catch to a type check maybe? Check for type array?

OfficialNoob: Well I think it would be a string.

OfficialNoob: JSON.parse would return a array

OfficialNoob:
@ajayyy I think I have added Map() support that should work.
Example: SB.config.sponsorTimes.set("x", true); (Synced)
Debug: SB.config.sponsorTimes.map (Local map)

ajayyy:
This is probably the best way to do it

ajayyy:
Can you please change your text editor to use 4 spaces?

ajayyy:
I think it is ready to merge now.

Next thing to do would be to replace some of [these messages](https://github.com/ajayyy/SponsorBlock/blob/master/content.js#L124) with a listener callback from SB.config. This probably should be done in another PR however to reduce merge conflicts with the invidious PR.

OfficialNoob:
Yeah new PR this one seems to have a lot of changes

PR #224, opened 2020-01-09
Added privacy confirmation for unlisted videos 

OfficialNoob:
Confirm if a user wants to use sponsor skip for unlisted videos.


ajayyy:
https://github.com/ajayyy/SponsorBlock/pull/264

ajayyy:
This is going to break for non English users. Switching to the new method should fix it.

PR #235, opened 2020-01-11
Formatting changes & proxy delete support

OfficialNoob:


ajayyy:
Please try to remember to use `let` and `const` when required.

OfficialNoob:
does const make much difference in a forEach

SB.js
```diff
@@ -99,18 +99,23 @@ function configProxy() {
...
+        get(obj, prop) {
             let data = SB.localConfig[prop];
             if(data instanceof Map) data = new MapIO(prop);
 
-			return obj[prop] || data;
+            return obj[prop] || data;
+        },
+	
+        deleteProperty(obj, prop) {
+	    chrome.storage.sync.remove(key);
```
ajayyy: This code was using `key` instead of `prop` and still working since `key` wasn't scoped properly. `key` was a global variable created by the for loop. I fixed it though.

ajayyy:
It doesn't matter for the `forEach` function since that is a function variable and is already scoped.

PR #243, opened 2020-01-12
Polish translation created

Deykun:


ajayyy:
Thanks! Would it be possible to get `fullName` under 50 characters? If not, I can just remove a space after the `-`.

Deykun:
>Thanks! Would it be possible to get `fullName` under 50 characters? If not, I can just remove a space after the `-`.

Hi.
Yes, sure i've updated it. 👍

Now `fullName` uses only 48 characters.

PR #244, opened 2020-01-13
Support options button when not in popup

OfficialNoob:


ajayyy:
This works?

OfficialNoob:
@ajayyy seems to work.

PR #245, opened 2020-01-16
Set options and help page to use html5

jplsek:
- Fix css when using html5
- Add encoding scheme to html
- Remove font-size change on title on hover. It's jarring
seeing the whole page move when trying to move your mouse
to the options.
- Remove center tags
- Remove unnecessary styles
- Fix some minor grammar

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -349,7 +349,7 @@
         "message": "Support Invidious"
     },
     "supportInvidiousDescription": {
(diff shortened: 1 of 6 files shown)


ajayyy:
Centering on the help page source code section and "enable invidious" button seems broken.

jplsek:
I just tested on master and clicking the enable invidious button brings me to the option page and the switch works fine (gives me the permission prompt).

ajayyy:
Oh sorry, I just meant the centering was broken. It is okay though, because I think I am going to remove it any way and just put an embed of the options page.

PR #247, opened 2020-01-16
Save Map content

OfficialNoob:


ajayyy:
Why did you remove it from delete?

OfficialNoob:
@ajayyy forgot to add it to the other delete

OfficialNoob:
I dont know what happend to the chrome.sync.remove

OfficialNoob:
I removed deleteProperty as it did not seem to be used for anything
The map can still be deleted with delete SB.config.mapname
And cleared with SB.config.mapname.clear()
Or to remove a key SB.config.mapname.delete(key)

PR #250, opened 2020-01-23
Add minimum duration option

Lartza:
This adds a new config option for minimum duration of sponsor segments to be skipped.

Option page location was decided based on the fact it takes precedence over anything else. Segments shorter than the setting cannot be skipped manually either.
I'm not sure if that's the best way but I also think it's the simplest way to please autoskip and manual skip users.

Having written the code I feel unsure about it but rather than bash my head on perfecting it alone I'd rather tweak or rewrite it if needed after some input from other people. Feel free to be as harsh as you want.

I've only tested this on Firefox.


options/options.js
```diff
@@ -68,6 +68,20 @@ async function init() {
...
+                let configValue = SB.config[numberChangeOption];
+                let numberInput = optionsElements[i].querySelector("input");
+
+                if (configValue != undefined) {
+                    numberInput.value = configValue;
+                }
+
+                numberInput.addEventListener("input", () => {
+                    SB.config[numberChangeOption] = numberInput.value;
+                });
```
ajayyy: Add a `break` statement here

options/options.js
```diff
@@ -68,6 +68,20 @@ async function init() {
...
             case "display":
                 updateDisplayElement(optionsElements[i])
+
+                break;
+            case "number-change":
+                let numberChangeOption = optionsElements[i].getAttribute("sync-option");
+                let configValue = SB.config[numberChangeOption];
+                let numberInput = optionsElements[i].querySelector("input");
+
+                if (configValue != undefined) {
```
ajayyy: It should never be undefined right? Because of the default.

Also, use `!==` here (I know my code doesn't always do that, but I am doing it in code from now on).

ajayyy:
It is probably best to ignore them when getting the sponsor times instead, since that would prevent the preview bar from displaying anything.

Lartza:
> 
> 
> It is probably best to ignore them when getting the sponsor times instead, since that would prevent the preview bar from displaying anything.

I do admit I didn't really think about that before. But if the segment isn't shown one could question that a sponsor isn't even submitted for that portion and that's why it wasn't skipped, not because of duration. Showing the bars doesn't seem that intrusive compared to that.

Joe-Dowd:
Yeah I'd still show the sponsor in the bar - anybody who's enabled the option will know that's why it's skipped, and they're more likely to see if it's an incorrect submission so they can report it.

Lartza:
> 
> 
> Yeah I'd still show the sponsor in the bar - anybody who's enabled the option will know that's why it's skipped, and they're more likely to see if it's an incorrect submission so they can report it.

Currently you don't get a notice from duration skipped sponsors though, which may or may not be bad for both knowing why a segment is not skipped or to report it easily.

But having the notice would mean refactoring the notice code a bit as in my mind it would also need an additional toggle so that you have more control over which notices to show. (Either because you want to be able to manual skip but not see notices about short sponsors or because you want to see notices about short sponsor ignores but not autoskips, etc)

ajayyy:
I see no reason to show a notice for nothing happening

Lartza:
The only reason I see would be to allow manual skipping but I don't see why that would be useful for that many people. (Set minimum duration to a relatively high value but still skip shorter than that but longish actual sponsor segments manually?)

ajayyy:
I have a feeling people will think it is broken if it does display, but not skip.

ajayyy:
> But if the segment isn't shown one could question that a sponsor isn't even submitted for that portion and that's why it wasn't skipped, not because of duration.

Maybe warn them if their submission is below the minimum they set?

Joe-Dowd:
> Currently you don't get a notice from duration skipped sponsors though, which may or may not be bad for both knowing why a segment is not skipped or to report it easily.
I didn't mean a notice (which I don't think should be sent) - I mean they'll know it's why it's set because they set the value in settings.

> The only reason I see would be to allow manual skipping but I don't see why that would be useful for that many people. (Set minimum duration to a relatively high value but still skip shorter than that but longish actual sponsor segments manually?)
Could be used by people that want to check if a sponsor report is incorrect to report/remove it. But imo if thats why they have the value set a boolean should skip setting would be a better call.

I'd personally prefer the overlay to be there but warning on submit would work if I forgot I'd set it and tried to submit something I saw.

Lartza:
But why a warning for length? Short sponsors are still valid sponsors. I guess we could store two lists of sponsor times and compare on submission but show the other one on the bar but that almost seems more trouble than what it's worth.

ajayyy:
A warning if they have the setting enabled as a reminder

ajayyy:
I think they should be ignored a entirely. No preview or anything

Lartza:
But should the submission then be outright blocked instantly until the length is over the user set minimum duration? We can't tell the user a sponsor time exists instantly in that spot afaik since it's a server response right? And the user doesn't know themself since the addon is hiding that information.

ajayyy:
Just a small warning message with `comfirm' can be used so they know they
are submitting something smaller than their minimum.

options/options.css
```diff
@@ -309,4 +309,13 @@ h1,h2,h3,h4,h5,h6 {
 
 svg {
     text-decoration: none; 
+}
+
+.number-container:before {
+    content: attr(label-name);
```
ajayyy: Is there a reason this is being set in css?

Lartza: It was copied from https://github.com/ajayyy/SponsorBlock/blob/b5996a9783d2b45993db3235efb6cdbbc864e0c3/options/options.css#L70

Though I do admit I was never 100% sure why it was set in css there either.

ajayyy:
Watch out when modifying arrays while looping through them. Bad things can happen. In this case, it would hide the wrong submission.

PR #257, opened 2020-02-02
SB.config bug fixes

OfficialNoob:
This should fix:

- Config values not getting decoded from JSON on page load

### DIFF
--- a/SB.js
+++ b/SB.js
@@ -153,7 +153,7 @@ function migrateOldFormats() { // Convert sponsorTimes format
 async function setupConfig() {
     await fetchConfig();
     addDefaults();
-    convertJSON();
+    decodeConfig();
     SB.config = configProxy();
     migrateOldFormats();
 }
@@ -184,8 +184,8 @@ function resetConfig() {
     SB.config = SB.defaults;
 };
 
-function convertJSON() {
-    Object.keys(SB.defaults).forEach(key => {
+function decodeConfig() {
+    Object.keys(SB.localConfig).forEach(key => {
         SB.localConfig[key] = decodeStoredItem(SB.localConfig[key], key);
     });
 }
--- a/content.js
+++ b/content.js
@@ -987,11 +987,9 @@ function sendSubmitMessage(){
 
                 submitButton.addEventListener("animationend", animationEndListener);
 
-                //clear the sponsor times
-                SB.config.sponsorTimes.delete(currentVideoID);
-
                 //add submissions to current sponsors list
-                sponsorTimes = sponsorTimes.concat(sponsorTimesSubmitting);
+                SB.config.sponsorTimes.set(currentVideoID, sponsorTimesSubmitting);
+                
                 for (let i = 0; i < sponsorTimesSubmitting.length; i++) {
                     // Add some random IDs
                     UUIDs.push(generateUserID());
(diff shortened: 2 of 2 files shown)


content.js
```diff
@@ -987,11 +987,9 @@ function sendSubmitMessage(){
...
                 submitButton.addEventListener("animationend", animationEndListener);
 
-                //clear the sponsor times
-                SB.config.sponsorTimes.delete(currentVideoID);
-
                 //add submissions to current sponsors list
-                sponsorTimes = sponsorTimes.concat(sponsorTimesSubmitting);
+                SB.config.sponsorTimes.set(currentVideoID, sponsorTimesSubmitting);
+                
                 for (let i = 0; i < sponsorTimesSubmitting.length; i++) {
```
ajayyy: This is when the times are cleared after a submit. Why are you adding them back?

SB.js
```diff
@@ -184,8 +184,8 @@ function resetConfig() {
     SB.config = SB.defaults;
 };
 
-function convertJSON() {
-    Object.keys(SB.defaults).forEach(key => {
+function decodeConfig() {
+    Object.keys(SB.localConfig).forEach(key => {
```
ajayyy: Oh shoot, how was this ever working?

OfficialNoob: If the key was used in SB.defaults it would still get decoded

content.js
OfficialNoob: Reverted

SB.js
ajayyy: Isn't the sponsorTimes in the defaults? Doesn't that mean that this shouldn't change the behavior?

ajayyy:
Is there still an issue related to sponsorTimes.concat?

OfficialNoob:
There was but it may have been because SB.config.sponsorTimes was not working

ajayyy:
https://github.com/ajayyy/SponsorBlock/issues/258

SB.js
OfficialNoob: It is in the defaults however there maybe something thats not.

OfficialNoob: I have been unable to get the error after this change maybe it only happens when the video has no times

OfficialNoob:
There could be an issue if a string was set as valid JSON

OfficialNoob:
@ajayyy should there be a type change prevention on the proxy?

This would: 
- allow custom objects or Maps to be detected easier and more reliable
- make config variables more predictable

The type could still be changed with SB.config.types.set("prop", "array")
```
function getName(obj) {
    return Object.prototype.toString.call(obj).replace(/^\[object (.+)\]$/,"$1").toLowerCase(); 
}
```

ajayyy:
@OfficialNoob yea, that's a great idea. But, there has to be a better way to do it then that.

ajayyy:
Also, is there a reason that the array is stored as a string? Chrome supports storing arrays

OfficialNoob:
@ajayyy because I dont want arrays to be converted to Maps
This would not be needed if I could get the type using `SB.config.types`

OfficialNoob:
`SB.config.types` would contain the initiation variable type such as from `SB.defaults`
Unless changed

ajayyy:
Check the typescript branch. Can we use typescript types to achieve this?

PR #261, opened 2020-02-04
Made decodeStoredItem detect item type

OfficialNoob:


src/config.ts
```diff
@@ -138,18 +138,19 @@ function encodeStoredItem(data) {
...
-        
-        if(!Array.isArray(str)) return data;
-        return new SBMap(id, str);
-    } catch(e) {
-
-        // If all else fails, return the data
-        return data;
+    if(!Config.localConfig[id]) return data;
+
+    if(Config.localConfig[id] instanceof SBMap) {
```
ajayyy: Won't it be an array?

OfficialNoob: Config.localConfig[id] should be the config value i think

ajayyy: Yea, but chrome stores SBMap as an array right (with decode)

OfficialNoob: Config.localConfig[id] is its old value

ajayyy: This is decoding it. Did you mean to use `Config.defaults`?

OfficialNoob: well it should be added when addDefaults() gets run.
however Config.defaults would probably be better to use

OfficialNoob: Is this resolved?

OfficialNoob:
@ajayyy its done.

ajayyy:
@OfficialNoob You can now test the config with `window.SB.config`. Tell me how it goes.

src/config.ts
```diff
@@ -138,18 +137,32 @@ function encodeStoredItem(data) {
...
+        try {
+            if (!Array.isArray(data)) return data;
+            return new SBMap(id, data);
+        } catch(e) {
+            console.error("Failed to parse SBMap: "+ id);
+        }
+    }
+
+    // This is the old format for SBMap (a JSON string)
+    if (typeof data === "string") {
```
OfficialNoob: I was trying to prevent this.
I don't want a string being converted to a map.

ajayyy: This is required to migrate the old data right?

OfficialNoob: Well it should be migrated after the version change
not on every read however I don't think any map needs to be migrated

ajayyy: It's easier to just migrate it this way. It can be removed in a few months.

OfficialNoob: It would be better to have it during the update process.
at least it will be fixed some day :D

ajayyy:
This has less chance of bugs occuring.

OfficialNoob:
Yeah as long as there never a string like [] for any key

OfficialNoob:
Tested current changes and it seems to work.
`SB.config` works now :D

ajayyy:
Looks like that migration code I added is not working since it is not working if `sponsorTimes` when updating and having left over config data.

ajayyy:
Seems good now I think

ajayyy:
@OfficialNoob Can you check it?

OfficialNoob:
@ajayyy tested it.

- Changes seem to result in the extension working :D
- SB.config seems to be no longer usable from the console.
- Enabling the skipper from the extension popup does not affect the current video without refresh. (Probably unrelated)

ajayyy:
@OfficialNoob It should work on the console on the background page and options page.

OfficialNoob:
Works fine at the options page

ajayyy:
This is for security reasons. Before, any extension with access to the YouTube page could grab the user id

OfficialNoob:
ok I think this is done now

ajayyy:
I'll merge it once I figure out what is causing the freezing issue

PR #285, opened 2020-02-23
Fix "0 sponsor" in popup

cherryblossom000:
The pop now says "So far, you've submitted 0 sponsor**s**" instead of "So far, you've submitted 0 sponsor".


ajayyy:
Can you change to !==

PR #292, opened 2020-03-07
Fix Russian translation and add missing strings

grishka:
Also fixed some typos in the English translation.

There are some problems I ran into:
* There's really no Russian word for "submission". I'm still not really sure what to call it, and this word is used rather extensively throughout the extension. I did my best.
* There are only two plural forms for quantities, like in English. In Russian, you need three for it to look natural. [See my explanation here](https://github.com/grishka/Smithereen/pull/6#issuecomment-595103350).
* Some parts aren't translatable — like this "since February" for example.
uBlock_Origin_-_Интернет-магазин_Chrome


ajayyy:
I plan to remove the "Since February" in some time, so I felt it wasn't worth translating.

ajayyy:
I am reading up on Russian plurals right now. That's so interesting how it works.

PR #301, opened 2020-03-20
Added restrictions to keybindings.

Joe-Dowd:
Setting certain keys as keybindings can cause it to interfere with keybindings youtube uses.

This restricts the user from setting those keys as keybindings, as well as any other keybinding the user has set SponsorBlock to listen for.

The method of alerting the user may need to be changed.

List of keys youtube uses taken from here:
https://www.hongkiat.com/blog/useful-youtube-keyboard-shortcuts-to-know/

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -320,6 +320,26 @@ function keybindKeyPressed(element: HTMLElement, e: KeyboardEvent) {
     var key = e.key;
 
     let button = element.querySelector(".trigger-button");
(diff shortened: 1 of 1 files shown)


Joe-Dowd:
Might need to remove Shift as a restricted keys, depending on some keyboard layouts.

Joe-Dowd:
I give up after testing (Shift is causing issues and I don't wanna touch it).

PR #315, opened 2020-04-04
Audio notification on skip

Joe-Dowd:
Options toggle switch to play an audio notification whenever a segment is skipped.
I made the ogg file myself so can provide the source file if needed.


src/options.ts
```diff
@@ -65,6 +65,10 @@ async function init() {
                                 showNoticeSwitch.checked = true;
                             }
 
+                            break;
+                        case "audioNotificationOnSkip":
+                            let audioNotificationOnSkip = <HTMLInputElement> document.querySelector("[sync-option='audioNotificationOnSkip'] > label > label > input");
+                            audioNotificationOnSkip.checked = Config.config[option];
```
ajayyy, line 71: Is there a specific reason you added this as a special case?

PR #350, opened 2020-05-13
Don't skip ads, hide controls & sponsor times preview bar when ads are playing

rafern:
Videos that start with a sponsor skip ads that occur before the video starts. This PR disables skipping when an ad (not banner ad) is playing and also hides the green sponsor preview bar and sponsor submission controls.

Example of video that always seems to start with a full player ad:
[https://www.youtube.com/watch?v=YuRYaTVSuhs](https://www.youtube.com/watch?v=YuRYaTVSuhs)
You can use that for verifying this works, if the ad at the start doesn't show, clearing cookies should make the ad show

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -24,7 +24,7 @@ var sponsorTimes: SponsorTime[] = null;
 //what video id are these sponsors for
 var sponsorVideoID: VideoID = null;
 
-// JSON video info
+// JSON video info 
 var videoInfo: any = null;
 //the channel this video is about
 var channelID;
@@ -40,6 +40,9 @@ var sponsorSkipped: boolean[] = [];
 
 //the video
 var video: HTMLVideoElement;
+
+/** The last time this video was seeking to */
+var lastVideoTime: number = null;
 
 var onInvidious;
 var onMobileYouTube;
@@ -121,7 +124,7 @@ var skipNoticeContentContainer: ContentContainer = () => ({
 
 //get messages from the background script and the popup
 chrome.runtime.onMessage.addListener(messageListener);
-
+  
 function messageListener(request: any, sender: any, sendResponse: (response: any) => void): void {
     //messages from popup script
     switch(request.message){
@@ -173,7 +176,7 @@ function messageListener(request: any, sender: any, sendResponse: (response: any
             // Start preview resetter
             if (previewResetter !== null){
                 clearTimeout(previewResetter);
-            }
+            } 
 
             previewResetter = setTimeout(() => previewResetter = null, 4000);
 
@@ -214,8 +217,8 @@ function messageListener(request: any, sender: any, sendResponse: (response: any
 
 /**
  * Called when the config is updated
- *
- * @param {String} changes
+ * 
+ * @param {String} changes 
  */
 function contentConfigUpdateListener(changes) {
     for (const key in changes) {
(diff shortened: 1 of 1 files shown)


ajayyy:
Thanks! What part of the code makes sure the preview bar reappears and skips are not rescheduled when the ad is over?

src/content.ts
```diff
@@ -464,6 +468,16 @@ function cancelSponsorSchedule(): void {
 function startSponsorSchedule(includeIntersectingSegments: boolean = false, currentTime?: number): void {
     cancelSponsorSchedule();
 
+    // Don't skip if advert playing and reset last checked time
+    if (document.getElementsByClassName('ad-showing').length > 0) {
```
ajayyy, line 472: Duplicate code?

src/content.ts
```diff
@@ -464,6 +468,16 @@ function cancelSponsorSchedule(): void {
...
+        // Reset lastCheckVideoTime
```
ajayyy, line 473: Why reset this?

src/content.ts
rafern: Good point, will test if nothing breaks with isAdPlaying instead, since i was only using it for UI

src/content.ts
rafern: Without this, if the sponsor is at the start, it won't be skipped after the ad is over, because it thinks that its a duplicate skip, hence why the last checked time is reset

src/content.ts
```diff
@@ -464,6 +468,16 @@ function cancelSponsorSchedule(): void {
 function startSponsorSchedule(includeIntersectingSegments: boolean = false, currentTime?: number): void {
     cancelSponsorSchedule();
 
+    // Don't skip if advert playing and reset last checked time
+    if (isAdPlaying) {
+        // Reset lastCheckVideoTime
+        lastCheckVideoTime = -1;
+        lastCheckTime = 0;
+
+        lastVideoTime = video.currentTime;
```
ajayyy, line 477: Shouldn't this be `lastVideoTime = null`?

Actually, it seems that variable is unused entirely and can probably just be removed.

src/content.ts
```diff
@@ -563,6 +577,8 @@ function sponsorsLookup(id: string) {
 
                 startSponsorSchedule();
             }
+
+            updateAdFlag();
```
ajayyy, line 581: Is this check required here? If so, it should be moved above the `startSponsorSchedule()` call, since it uses the variable.

ajayyy:
Can you make `isAdPlaying` reset in the `resetValues()` function, as that will probably prevent some edge case bugs.

src/content.ts
rafern: I thought the variable was used since it was also in the pause event listener, but it turns out its unused so its removed now

src/content.ts
rafern: Its required because it detects whether it changed from an ad playing or video playing. Its moved above startSponsorSchedule now

rafern:
Done, isAdPlaying will reset in resetValues

ajayyy:
Thanks a lot, sorry about asking for so many changes

ajayyy:
I hit merge too early, here is the commit I missed for future reference: 178b122ab8e722799311c68c33c06a4f13be1423

PR #389, opened 2020-06-22
Fix typo in Chrome Users badge

kittenparry:
Chome -> Chrome


ajayyy:
Not sure how I never noticed, thanks!

PR #395, opened 2020-06-26
Added config option for excluding skipped content from the video duration

NDevTK:
Config.config.timeWithSkips
Config.config.hideRealTime

### DIFF
--- a/public/_locales/en/messages.json
+++ b/public/_locales/en/messages.json
@@ -289,6 +289,12 @@
     "audioNotificationDescription": {
         "message": "Audio notification on skip will play a sound whenever a sponsor is skipped. If disabled (or auto skip is disabled), no sound will be played."
     },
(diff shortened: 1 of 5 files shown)


ajayyy:
Looks good, thanks! This just changes the total time right? Not the current time. It might be confusing if the current time is over the total. 

Maybe it would look nice to have the time in brackets? `12:09 (10:20)`. Then, it probably could even be enabled by default.

NDevTK:
@ajayyy it does not affect the current time.
I have added support for stuff like "12:09 (10:20)" when `Config.config.hideRealTime` is false

NDevTK:
@ajayyy any other changes?

ajayyy:
I think switching to using the actual sponsorTimes variable would be nice, but not required. Then the config options need to be added to the options page.

PR #438, opened 2020-08-13
Dependencies updates, reload tip and .editorconfig added

MRuy:


ajayyy:
I'd rather keep jest there for now as a reminder to make me add some tests at some point

MRuy:
> I'd rather keep jest there for now as a reminder to make me add some tests at some point

Thats why i changed the TEST script in the package.json to this:  

`"test": "echo \"Error: no test specified\" && exit 1"`

PR #440, opened 2020-08-19
Design change of popup.html

mbledkowski:
Hi, I have changed design for popup window. 
If something is wrong with my changes, I can correct those.

New design:
Image of new design
Old design:
Image of old design


ajayyy:
Looks great, great job!

I think we have to do something else when they don't have a username yet. It looks really ugly in the default state. Also, the popup should probably have a max width and have word wrap enabled. I'm also not sure why the whitelist button is the wrong colour.

image

Would you be able to put that whitelist warning somewhere when someone clicks "Add to whitelist"?

Here are some notes for discussion:

Missing things:

- "Segment starts", "Segment ends"
    - This is the only way to submit on invidious
- Voting

mbledkowski:
Hi, I am very sorry that it took so long. Missing things that you mentioned have been fixed.

ajayyy:
Great! I'll check it out soon, thanks a lot!

mbledkowski:
Hi, is there something I should change, and are you planning to merge this pull request?

ajayyy:
Nope, but I plan to do some testing, so I have to wait a bit until I have some more time

ajayyy:
I made some small changes.

Here are some other changes I was thinking about:

- Change voting to have voting buttons appear below the times instead of to the right

- Make username textbox take the whole space when editing
   - The submissions number could disappear maybe?
   - This is because of the username being very long by default

ajayyy:
image

Current look

ajayyy:
Also, it seems broken on the YouTube page when light theme is on. I think it's just that the background colour is not working.

mbledkowski:
> Also, it seems broken on the YouTube page when light theme is on. I think it's just that the background colour is not working.

It looks ok for me
image

mbledkowski:
Hi, are you planning to merge those changes? Should I change something?

ajayyy:
Hey, yes I wanted these changes before merging https://github.com/ajayyy/SponsorBlock/pull/440#issuecomment-708863002. I was planning on doing them myself but have been very busy.

ajayyy:
Thanks again for your work! Sorry it took so long to merge.

PR #459, opened 2020-09-03
Added hash prefix implementation (server code not yet merged) 

Joe-Dowd:
This feature changes the way segments are obtained from the server in a way that improves user privacy, but disables fetching sgements while the video is being played should they be submitted. 

Instead of requesting segments using the video ID (the string after watch?v=), the video ID is hashed and the first 4 characters of the hash are sent to the server, which returns all videos where the hash also starts with those characters. The client then selects the video it wants (if any) and uses the segments provided in that request.

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -633,8 +633,6 @@ function sponsorsLookup(id: string) {
     getRequest.then(async (response: FetchResponse) => {
         if (response?.ok) {
             let getResult = JSON.parse(response.responseText);
-            console.log(getResult);
-            alert(getResult.length);
             if (Config.config.hashPrefix) {
                 getResult = getResult.filter((video) => video.videoID === id);
                 if (getResult.length > 0) {
--- a/src/content.ts
+++ b/src/content.ts
@@ -632,20 +632,23 @@ function sponsorsLookup(id: string) {
     }
     getRequest.then(async (response: FetchResponse) => {
         if (response?.ok) {
-            let getResult = JSON.parse(response.responseText);
+            let result = JSON.parse(response.responseText);
             if (Config.config.hashPrefix) {
-                getResult = getResult.filter((video) => video.videoID === id);
-                if (getResult.length > 0) {
-                    getResult = getResult[0].segments;
-                    if (getResult.length === 0) { // return if no segments found
+                result = result.filter((video) => video.videoID === id);
+                if (result.length > 0) {
+                    result = result[0].segments;
+                    if (result.length === 0) { // return if no segments found
+                        retryFetch(id);
                         return;
                     }
                 } else { // return if no video found
+                    console.log("refetching")
+                    retryFetch(id);
                     return;
                 }
             }
 
-            let recievedSegments: SponsorTime[] = getResult;
+            let recievedSegments: SponsorTime[] = result;
             if (!recievedSegments.length) {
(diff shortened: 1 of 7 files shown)


Joe-Dowd:
https://github.com/ajayyy/SponsorBlockServer/pull/127

src/content.ts
```diff
@@ -618,12 +618,36 @@ function sponsorsLookup(id: string) {
...
+    } else {
+        getRequest = utils.asyncRequestToServer('GET', "/api/skipSegments", {
+            videoID: id,
+            categories
+        });
+    }
+    getRequest.then(async (response: FetchResponse) => {
         if (response?.ok) {
-            let recievedSegments: SponsorTime[] = JSON.parse(response.responseText);
+            let getResult = JSON.parse(response.responseText);
```
ajayyy, line 635: Maybe the variable from `getResult` to `result`?

src/content.ts
```diff
@@ -618,12 +618,36 @@ function sponsorsLookup(id: string) {
...
+            if (Config.config.hashPrefix) {
+                getResult = getResult.filter((video) => {
+                    return video.videoID = id;
+                });
+                if (getResult.length === 1) {
```
ajayyy, line 640:
```suggestion
                if (getResult.length > 0) {
```

src/utils.ts
```diff
@@ -378,6 +379,20 @@ class Utils {
...
+
+        for (let i = 0; i < times; i++) {
+            let hash = sha256.create();
+            hash.update(value);
+            hash.hex();
+            value = hash.toString();
+        }
+
+        return value;
+    }
```
ajayyy, line 394: This is great to add, as we'll be able to pre-calculate the public userID on the frontend now.

ajayyy (changes requested):

ajayyy:
On my local environment, segments are loading very slowly (Server-side).

ajayyy:
Index fixed it

src/utils.ts
```diff
@@ -378,6 +379,20 @@ class Utils {
     isFirefox(): boolean {
         return typeof(browser) !== "undefined";
     }
+
+    getHash(value: string, times=5000): string {
+        if (times <= 0) return "";
+
+        for (let i = 0; i < times; i++) {
```
xPaw, line 386: What is the idea behind hashing it 5k times?

And take a look at native function: https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest

Joe-Dowd: That's a default value taken from the server code 
https://github.com/ajayyy/SponsorBlockServer/blob/master/src/utils/getHash.js#L3-L12

On the server side, userIDs are hashed 5000 times before being stored in the database.

ajayyy: It is to be the same as the server, which hashes 5k times for it's operation.

That's a good idea to use the browser crypto function, thanks!

xPaw: But what is the reason for hashing it more than once?

ajayyy: It was to make brute forcing harder since the hash of the private userID is in the public database.

xPaw: Fair enough, I didn't spot that it's called with times=1 for video IDs.

PR #486, opened 2020-09-26
Fixing typo and removing 'anglisismes'

SirCipherz:
Changing timer to minuteur : it's an anglicisme, not a french word, same for liker and follow
Changing "segment non-musicale" to "segment non-musical" because segment is a masculine word
And fixing a typo at line 499


ajayyy:
Thanks!

PR #490, opened 2020-10-01
fix: resolve translation template issue

### LINKED ISSUE #447: Translation problem
I've noticed a problem with the translation to german.
image
"Überspringen Unterbrechung?" should be "Unterbrechung überspringen?" but on crowdin I could only find these strings as separate parts so the grammar cannot be fixed in the translation. You would need to update the translation template to include all whole strings and not piece together the translation as it is doing now.

skyfrk:
Hi there! This is my attempt to resolve #447.

### The problem (citing from #447)

> "Überspringen Unterbrechung?" should be "Unterbrechung überspringen?" but on crowdin I could only find these strings as separate parts so the grammar cannot be fixed in the translation. You would need to update the translation template to include all whole strings and not piece together the translation as it is doing now.

### Solution

I updated the default locale translation template with new strings, translated them into German and updated the skip notification popup code to use them. I hope that this is the way to go - I didn't work with crowdin yet 😬.

### Alternative solution

An alternative solution may add locale specific code [here](https://github.com/ajayyy/SponsorBlock/blob/56136d598bc1c960bc50ab09bf21ead903010370/src/components/SkipNoticeComponent.tsx#L80):

```js
if (!this.autoSkip) {
    noticeTitle = chrome.i18n.getMessage("skip") + " " + categoryName + "?";

    if(chrome.i18n.getUILanguage() == "de"){
        noticeTitle = categoryName + " " + chrome.i18n.getMessage("skip").toLowerCase() + "?";
    }
}
```

### DIFF
--- a/public/_locales/de/messages.json
+++ b/public/_locales/de/messages.json
@@ -619,5 +619,41 @@
     },
     "unsubmittedWarningDescription": {
         "message": "Zeige eine Benachrichtigung an, wenn Sie ein Video mit nicht hochgeladenen Segmenten verlassen"
+    },
+    "skip_multiple_segments": {
+        "message": "Mehrere Segmente überspringen?"
+    },
+    "skip_category_sponsor": {
+        "message": "Sponsor überspringen?"
+    },
+    "skip_category_intro": {
+        "message": "Unterbrechung/Intro Animation überspringen?"
+    },
+    "skip_category_intro_short": {
+        "message": "Unterbrechung überspringen?"
+    },
+    "skip_category_outro": {
+        "message": "Endkarten/Credits überspringen?"
+    },
+    "skip_category_interaction": {
+        "message": "Interaktions-Erinnerung (Abonnieren) überspringen?"
+    },
+    "skip_category_interaction_short": {
+        "message": "Interaktions-Erinnerung überspringen?"
+    },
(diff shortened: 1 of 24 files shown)


ajayyy:
Maybe this would be better with the string being changed to something like "Skip {1}" and then doing String.replace, but then all the translations would have to be updated manually.

skyfrk:
@ajayyy Neat idea! This way there is no redundancy in the translation files. I'll update the PR.

skyfrk:
@ajayyy Ready for review 🎉

ajayyy:
Looks good, thanks!

Before merging, this will have to be fixed in all translations

skyfrk:
@ajayyy Ok, I'll add the strings!

ajayyy:
I think it may be easiest to make a script to do this automatically. I'll try to do this when I have some time.

skyfrk:
@ajayyy I added `skip_category` strings to all languages which have the `skip` string defined. Is this ok?

skyfrk:
@ajayyy what do you think? :D

ajayyy:
Looks good! Did you use a script to do this, or did you do it manually?

skyfrk:
It was manual labor 🙈

PR #502, opened 2020-10-06
Fix: skipping to the end of looped videos #426

### LINKED ISSUE #426: Seeking to the last second moved on to the next video in the playlist even if loop is enabled
__Steps to reproduce:__

* Play this link: https://www.youtube.com/watch?v=cSnkWzZ7ZAA&list=PLgDP5UKLWaCA_cJeIJBcVLbLFid49AMQh&index=11
* Right click & enable loop on video
* At the end there is a non-music skip & I've non-music skip enabled in SB settings..

Expected:
* Song should play again after the non-music skip, since Loop is enabled 😋

Actual:
* Next song in the playlist starts playing! No no no! 😢

SuspiciousActivity:
This should fix #426 

**The problem:**
If SponsorBlock skips to the end of a video in a playlist which is set to loop, the looping doesn't work, instead it goes to the next video in the playlist.

**Solution:**
Check if the time to skip to is at the end of the video, if it is and the video is set to loop, jump to the beginning.

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -1007,7 +1007,13 @@ function skipToTime(v: HTMLVideoElement, skipTime: number[], skippingSegments: S
     let autoSkip: boolean = utils.getCategorySelection(skippingSegments[0].category)?.option === CategorySkipOption.AutoSkip;
 
     if ((autoSkip || sponsorTimesSubmitting.includes(skippingSegments[0])) && v.currentTime !== skipTime[1]) {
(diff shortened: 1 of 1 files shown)


ajayyy:
Looks great, thanks!

MathisSchm:
Should it work already? Because it doesn't

ajayyy:
When the next update gets pushed

PR #521, opened 2020-10-21
getYouTubeVideoID() - support youtube-local pathnames

zrose584:
https://github.com/user234683/youtube-local/issues/30

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -801,17 +801,14 @@ function getYouTubeVideoID(url: string) {
     }
 
     //Get ID from searchParam
-    if (urlObject.searchParams.has("v") && ["/watch", "/watch/"].includes(urlObject.pathname) || urlObject.pathname.startsWith("/tv/watch")) {
+    let m = null;
+    if (urlObject.searchParams.has("v") && urlObject.pathname.match(/^\/((youtube\.com\/)?watch\/?$|tv\/watch)/)) {
         let id = urlObject.searchParams.get("v");
         return id.length == 11 ? id : false;
-    } else if (urlObject.pathname.startsWith("/embed/")) {
-        try {
-            return urlObject.pathname.substr(7, 11);
-        } catch (e) {
-            console.error("[SB] Video ID not valid for " + url);
-            return false;
-        }
-    } 
+    }
+    else if (m = urlObject.pathname.match(/^\/(?:youtube.com\/)?(?:embed|youtu.be)\/([^/]{11})/)) {
+        return m[1];
+    }
     return false;
 }
 


ajayyy:
Holding off on merging this until I decide how to change the UI to indicate that more than just Invidious instances can be added.

zrose584:
I think the current MR is already valuable for those who know this feature/functionality is present.
So staling this PR because of UI-Decisions seems a bit unfortunate..

ajayyy:
This breaks for YouTube tv URLs: 

`https://www.youtube.com/tv/#/watch?resume&v=dQw4w9WgXcQ`

src/content.ts
```diff
@@ -801,17 +801,14 @@ function getYouTubeVideoID(url: string) {
     }
 
     //Get ID from searchParam
-    if (urlObject.searchParams.has("v") && ["/watch", "/watch/"].includes(urlObject.pathname) || urlObject.pathname.startsWith("/tv/watch")) {
+    let m = null;
+    if (urlObject.searchParams.has("v") && urlObject.pathname.match(/^\/((youtube\.com\/)?watch\/?$|tv\/watch)/)) {
```
ajayyy, line 805: What is `(youtube\.com\/)?` for?

src/content.ts
```diff
@@ -801,17 +801,14 @@ function getYouTubeVideoID(url: string) {
...
-    } else if (urlObject.pathname.startsWith("/embed/")) {
-        try {
-            return urlObject.pathname.substr(7, 11);
-        } catch (e) {
-            console.error("[SB] Video ID not valid for " + url);
-            return false;
-        }
-    } 
+    }
+    else if (m = urlObject.pathname.match(/^\/(?:youtube.com\/)?(?:embed|youtu.be)\/([^/]{11})/)) {
```
ajayyy: What is `youtu.be` for?

src/content.ts
zrose584: youtube-local routes `https://youtube.com/watch?v=<vid>` to `http://your-domain.org/youtube.com/watch?v=<vid>`
It does this because it has to route multiple domains, and it probably makes things simpler.

So if youtube-local were to soon also support another video platform (e.g. dailymotion), and they also use /watch?v=<vid>, this addon will correctly not transmit the id.
Also note that only `^\/watch` would not match youtube-local's routes.

src/content.ts
zrose584: see above

ajayyy: `youtu.be` should never be a valid URL since it is always redirected

zrose584: I just tested it, and youtube-local does not redirect the `/youtu.be/<video_id>` path to `/youtube.com/watch?v=video_id`.
It just serves the video as if `/youtube.com/watch?v=video_id` was used.
Two ways:
  * Merge it as is. The allowance of `/youtu.be/<video_id>` is not "hurting" anoyne, right?
  * Remove `youtu.be` from the regex, and open a ticket on youtube-local. I believe that it should be a quick fix, but I am not 100% sure.. Maybe not being redirected is a feature (idk)?

gorwhh (approved):

user12345678905:
Please, complete this pull request, people want block not only sponsor segments!

war59312:
If not going to merge these pull requests really should just close them all.

PR #536, opened 2020-11-06
Use formatted time inputs for segment editing

### LINKED ISSUE #535: Time input box incorrect decimal separator
The second time input box where you correct the skip point show a dot for me and not a comma as a decimal separator. The first one is correct localized.

image

opl-:
Changes the segment editor to use a single input field for both the minutes and seconds part of the segment boundary times.

Additionally, removes unused util functions and saves the edits on changes to allow users to preview their changes immediately on the progress bar and to allow previewing by manually clicking on the progress bar without accidentally using the old values.

If either of the fields contains an invalid input, the time change is ignored entirely.

image

The input field is rather permissive, allowing for quick inputs:

- The already used time format is supported (`1:20:34.12` = 1h20m34s120ms)
- An empty field is interpreted as "start of the video" (0s)
- Input containing just a number is interpreted as seconds (`24` = 24s, `15.3` = 15s300ms)
- Input containing just a number followed by `:` is interpreted as minutes (`1:` = 1m)
- Values higher than expected are allowed and interpreted correctly (`1:65` = 2m05s)
- Using a comma instead of a period is allowed, which is helpful for locales that use it as the decimal separator (fixes #535)

### DIFF
--- a/public/content.css
+++ b/public/content.css
@@ -372,12 +372,8 @@ input::-webkit-inner-spin-button {
 	-moz-appearance: textfield;
 }
 
-.sponsorTimeEditMinutes {
-	width: 30px;
-}
-
-.sponsorTimeEditSeconds {
-	width: 60px;
+.sponsorTimeEditInput {
+	width: 90px;
 }
 
 .sponsorNowButton {
--- a/src/components/SponsorTimeEditComponent.tsx
+++ b/src/components/SponsorTimeEditComponent.tsx
@@ -20,7 +20,7 @@ export interface SponsorTimeEditProps {
 
 export interface SponsorTimeEditState {
     editing: boolean;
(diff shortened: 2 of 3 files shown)


opl-:
Realized today that the regex was only capturing a single digit for both the hours and minutes. No idea how I managed to make that mistake but it's fixed now.

ajayyy:
Cool idea! Does it break if someone adds other characters or extra colons? What happens if someone tried to add `1:02:30.371`?

opl-:
Extra characters anywhere (like letters or spaces) will cause it to reject the input and use the last valid combination of start and end. I realized I didn't mention this in the comment, but hours are also accounted for in the format, meaning up to two `:` are allowed (`1:02:30.371` = 1h2m30.371s).

PR #549, opened 2020-11-21
Clean up Preview Bar

### LINKED ISSUE #517: Preview bar draw order
Shorter segments should be drawn last (at the top)

opl-:
Large commit overhauling most of the preview bar code.

Fixes:
- Segments hidden by longer segments (fixes #517)
- Duration with skips not accounting for segment overlaps
- Duration with skips not accounting for user's skip choices
- Segment category text in preview tooltip overlaps the seek bar
- Segment category text in preview tooltip breaks for timestamps over one hour
- `previewBar.ts` lacks function argument and return types
- Label on the progress bar tooltip not cleaned up on remove
- General code style issues


src/content.ts
```diff
@@ -819,46 +821,58 @@ function getYouTubeVideoID(url: string) {
...
+                category: segment.category,
+                preview: false,
+            });
+        });
     }
 
-    previewBar.set(utils.getSegmentsFromSponsorTimes(allSponsorTimes), types, video.duration)
+    sponsorTimesSubmitting.forEach((segment) => {
+        previewBarSegments.push({
+            timestamps: segment.segment as [number, number],
```
ajayyy, line 856: To be consistent, this should probably be `segment`

ajayyy (changes requested):

src/content.ts
opl-: I believe I named it `timestamps` instead of `segment` because the type itself is already called `Segment`. This results in slightly clunky naming such `segment.segment`. I also didn't want to name the `PreviewBarSegment` type `PreviewSponsorTime` even though it would be more consistent, because the name just seems awkward and confusing.

That said, this does make the naming within the project less consistent and perhaps this isn't the PR to fight it. Thoughts?

ajayyy: I agree with you. I'll fix the others and commit to your branch

ajayyy: So, it seems the old naming kind of has to stay since unsubmitted segments are stored using it and the API still uses it. The name "sponsorTime" at least should be able to change without issues though

src/content.ts
```diff
@@ -819,46 +821,58 @@ function getYouTubeVideoID(url: string) {
...
+    });
+
+    previewBar.set(previewBarSegments, video.duration)
 
     if (Config.config.showTimeWithSkips) {
-        showTimeWithoutSkips(allSponsorTimes);
+        const skippedSegments = previewBarSegments.filter((segment) => {
+            // Count the segment only if the category is autoskipped
+            return utils.getCategorySelection(segment.category)?.option === CategorySkipOption.AutoSkip;
+        });
```
ajayyy, line 868: Discussion here:   [Matrix](https://matrix.to/#/!CocoOGqmUMccHzIplA:blob.cat/$RlmER-19cusXPjUcmisuEvCSvZfZiVZOPsaVnGYqsjU?via=blob.cat&via=matrix.org&via=privacytools.io)/[Discord](https://discord.com/channels/603643120093233162/603643299961503761/788110191786197002)

ajayyy (changes requested):

ajayyy:
I just noticed that your other pull request will have conflicts now due to my refactor. Maybe we can just undo my commits and do that later

opl-:
Yeah. If you don't mind I'll just reset the branch back to avoid confusion and you can deal with the refactor later, ok?

ajayyy:
Seems to be an issue with the hover text

broken hover text

The same issue occurs when I checkout your commit, so I don't think it is caused by my changes

opl-:
I've been running this code for over a month now and haven't seen that issue. Not sure what's causing it for you, so you'll have to investigate yourself. Feel free to contact me on Discord about it.

The lines responsible for toggling the class are:

https://github.com/ajayyy/SponsorBlock/pull/549/files#diff-92cafe970f45d808f373350ae52000446b48dd5749a6cb4a5c7f1032e3b0581eR116-R118

The related styles are here:

https://github.com/ajayyy/SponsorBlock/pull/549/files#diff-455e53e45015aaf4f78b2555eae498aa5cc677e652604ca508af4ccb46ed58faR21-R23

opl-:
It finally happened to me today. I assume YouTube styles changed somehow and started overriding my styles. Either that or it was dependent on the order of the styles in the DOM or something. Regardless, adding an `!important` fixes it and shouldn't affect anything since it's specifically for hiding a single component of ours.

ajayyy:
Thanks!

There still seems to be a layout issue when there are not YouTube chapters. I'm not sure what is causing it.

Current version:
image

This PR:
image

opl-:
Yup, I seem to be finding a bunch of random issues now too. I have no idea what happened, I swear it was working when I tested it earlier. I'll look into this later. Initial guess is that it's more issues with CSS rule ordering, but that's just a guess.

ajayyy:
Yea, I guess we might have to put a lot of `!important` tags to prevent YouTube updates from messing with things.

PR #564, opened 2020-12-12
ESLint

FoseFx:
First of all, sorry this PR will be massive.

This PR adds eslint, @typescript-eslint/eslint.plugin and eslint-plugin-react with only defaults.
I did not change any logic for this and only applied trivial lint fixes.
Because I dont want the lint to fail I overwrote a couple of rules to warn instead of error.

I'll take care of them in follow up PRs.
While I did add a `npm run lint` and `npm run lint:fix` script the CI does not use it yet, I believe
this should be done in another follow up.


FoseFx:
I'll wrap things up here and fix all the other linting warnings in future PRs.

ajayyy:
I didn't know about `unknown` type, that's cool. Thanks!

PR #569, opened 2020-12-14
Clean up segment creation code

opl-:
Noteworthy changes:

- Adds ability to cancel creating a segment (uses the "upload failed" icon)
- Makes segment creation fully the responsibility of the content script, with the popup script buttons simply doing RPC
- Adds types to the `Utils.wait` function
- Fixes segment timestamps backwards if user marks segment end earlier than the start


ajayyy:
This causes the popup to immediatly disappear when you try creating a segment in the popup.

opl-:
Just tested that on Firefox and... It doesn't. The popup stays open for me. What browser were you running when this happened?

opl-:
Rebased onto latest master. I looked it over, did some testing, and didn't spot any new issues. It is a large change though, so you should probably review it with a fresh eye regardless.

ajayyy (changes requested):
I'm still getting the popup issue on both chrome and firefox. The browser popup menu is working, but the in-page menu (info button), is closing when buttons are pressed.

opl-:
Ah, I see. I thought you were talking about the browser popup and not the "info menu", as the code calls it. I missed that since I usually have that button completely disabled. Reason it closes is because of this code that I "inherited":

https://github.com/ajayyy/SponsorBlock/blob/5a861f97724f47eced6fee18feb6bb50a2a91ffa/src/content.ts#L1162-L1164

At first I fixed this by adding an argument to some functions, but I realized that it'd be cleaner to create a method for it, so I did that. This should make it easy to get rid of the workaround later when we can notify the popup of data changes.

ajayyy (approved):

PR #592, opened 2020-12-23
English: inclusive wording

Madis0:
Replaced references of "whitelist" with "allowlist". Did not currently refactor the code.


ajayyy:
I don't think allowlist makes sense in this context, and the plan is to revamp this to use "channel overrides" instead of a strict whitelist in the future anyway

PR #598, opened 2020-12-24
Enforce .editorconfig using eslint

FoseFx:
Sets "linebreak-style", "eol-last" and "indent" in `.eslintrc.js` according to `.editorconfig` and applies automatic fix for consistency in the code-base.


ajayyy:
Let's run this once those other pull requests are merged

PR #599, opened 2020-12-24
Replace self-equals-this code with arrow functions

FoseFx:
This should mark the end of the refactor efforts related to eslint! :tada:


ajayyy:
Ah, nice! I didn't know about that

PR #632, opened 2021-01-28
New Crowdin updates

ajayyy:

PR #637, opened 2021-02-03
Fix German pronouns and typo

catearcher:
This PR fixes two problems in the German translation of the `supportInvidiousDescription` string.

The rest of the file uses non-honorific pronouns, so this line should also use them (i.e. "du" instead of "Sie"). See also [my comment in PR #632](https://github.com/ajayyy/SponsorBlock/pull/632#pullrequestreview-582290744).

Plus: There is a typo in the word "Incongnito" (an extra "N"), and it is linked to the word "Modus" which should be uppercase when linked with a hyphen. But even if none of those typos were there it would *still* be incorrect, because Chrome uses the spelling "Inkognitomodus" (which is what I changed it to in this PR).

### DIFF
--- a/public/_locales/de/messages.json
+++ b/public/_locales/de/messages.json
@@ -348,7 +348,7 @@
         "message": "Invidious-Kompatibilität"
     },
     "supportInvidiousDescription": {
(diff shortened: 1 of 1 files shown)


ajayyy:
Thanks!

PR #643, opened 2021-02-07
Fixed getFormattedTime sometimes returning (NaN:NaN)

gubareve:
getFormattedTime() with nothing passed into it and a couple other edge cases returns (NaN:NaN). This is a problem since on a slow machine getFormattedTime gets called without anything passed into it around the same time the computer is fetching the sponsor data. I haven't found the exact time/cause/scenario that it is causing it to return  this, but this function at some point returns (NaN:NaN), and this will definitely fix it. I've have a slow machine so it's really been bugging me seeing (NaN:NaN) in the video bar for a couple seconds before the page fully loads. Hopefully this should fix it.

P.S. If you want some other case during the loading (e.g. nothing instead of "0:00") that would work as well.

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -1649,5 +1649,5 @@ function showTimeWithoutSkips(skippedDuration: number): void {
     
     const durationAfterSkips = utils.getFormattedTime(video.duration - skippedDuration)
 
-    duration.innerText = (durationAfterSkips == null || skippedDuration <= 0) ? "" : " (" + durationAfterSkips + ")");
+    duration.innerText = (durationAfterSkips == null || skippedDuration <= 0) ? "" : " (" + durationAfterSkips + ")";
 }
--- a/src/content.ts
+++ b/src/content.ts
@@ -1649,5 +1649,5 @@ function showTimeWithoutSkips(skippedDuration: number): void {
     
     const durationAfterSkips = utils.getFormattedTime(video.duration - skippedDuration)
 
-    duration.innerText = durationAfterSkips === "" ?  (skippedDuration <= 0 ? "" : " (" + durationAfterSkips + ")") : "";
+    duration.innerText = (durationAfterSkips == null || skippedDuration <= 0) ? "" : " (" + durationAfterSkips + ")");
 }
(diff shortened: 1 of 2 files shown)


ajayyy:
Nice catch! I think it might be better to just show nothing in this case.

gubareve:
@ajayyy How about this?

src/utils.ts
```diff
@@ -376,6 +376,9 @@ class Utils {
             //add a zero
             minutesDisplay = "0" + minutesDisplay;
         }
+        if (isNaN(hours) || isNaN(minutes)) {
+            return ""
+        }
```
ajayyy, line 381:
```suggestion
        if (isNaN(hours) || isNaN(minutes)) {
            return null;
        }
```

src/content.ts
```diff
@@ -1646,6 +1646,8 @@ function showTimeWithoutSkips(skippedDuration: number): void {
 
         display.appendChild(duration);
     }
+    
+    const durationAfterSkips = utils.getFormattedTime(video.duration - skippedDuration)
 
-    duration.innerText = skippedDuration <= 0 ? "" : " (" + utils.getFormattedTime(video.duration - skippedDuration) + ")";
+    duration.innerText = durationAfterSkips === "" ?  (skippedDuration <= 0 ? "" : " (" + durationAfterSkips + ")") : "";
```
ajayyy, line 1652: It seems the logic here is reversed.

```suggestion
    duration.innerText = (durationAfterSkips == null || skippedDuration <= 0) ? "" : " (" + durationAfterSkips + ")");
```

ajayyy (changes requested):

src/utils.ts
```diff
@@ -357,7 +357,7 @@ class Utils {
         });
     }
 
-    getFormattedTime(seconds: number, precise?: boolean): string {
+    getFormattedTime(seconds: number, precise?: boolean): string | null {
```
ajayyy, line 360: I don't think this is needed

ajayyy:
Thanks!

Issue #655, opened 2021-02-22
Add Name for userID Field in Seetings so Password Managers can autofill


DrDragonKiller:
I propose a tiny improvement for the extension website where the user changes settings:
There is this input field for changing the userID. When inspecting the element you see it only has a class and a type. 
https://github.com/ajayyy/SponsorBlock/blob/d2f377c8d745c5c8f4e2ede2ea7b930cb4d1c670/public/options/options.html#L375
This means it's not possible to identify only on these attributes alone.
Adding an additional  attribute like name or id would allow for its identification.
This would allow password managers (like bitwarden) to autofill this field.

Will be implemented in #656

DrDragonKiller:
see issue: not possible apparently

PR #656, opened 2021-02-22
Added Name Attribute to userID Input Field

DrDragonKiller:
Implements #655

### DIFF
--- a/public/options/options.html
+++ b/public/options/options.html
@@ -372,7 +372,7 @@
 				<div class="option-hidden-section hidden">
 					<br/>
 	
-					<input class="option-text-box" type="text">
+					<input class="option-text-box" type="text" name="userID">
 	
 					<br/>
 					<br/>


ajayyy:
I'm not sure if it is a good idea to "encourage" this field to be user-created, since it must be randomly generated, as it is not salted.

DrDragonKiller:
when you install the addon the first time it will generate a random id for you. Also the option of changing is collapsed by default.
So I don't think anyone will craft an id by hand (maybe this could be added to the description / it could be salted).

My usecase was, that I want to compare stats and thus change my account sometimes. Currently I always have to copy the long id and paste it there. This PR would allow me to let bitwarden autofill the id in the field whenever I want it to.

ajayyy:
Have you tested if Bitwarden is able to access this page? From a quick search, it doesn't look like extensions can access these pages. 

uBlock does seem to be able to detect the page url, but "block element" isn't working for me.

DrDragonKiller:
you're right. I thought this limitation was only for modifying the page. (Okay technically inpouting stuff also modifies the page but I thought this would work)
I tried setting the name and tried autofill and it didn't work, neither did LastPass recognized any changed fileds. I also tried on the Tamper Monkey Extension Page - No luck there either with both.

Can be closed as it appears to not be possible.

ajayyy:
Thanks for looking into it!

PR #657, opened 2021-02-22
Fix options button icon in firefox

aaaeka:
The options icon before was not vertically centered. This only appeared in firefox.
comparison


ajayyy:
Hey, can you verify that this doesn't break when you click whitelist channel?

aaaeka:
Sure.

fix

PR #666, opened 2021-03-04
Fix the videoRoot to also work on invidious embeds

### LINKED ISSUE #645: Support for invidious embed links
Sponsor block does not work for embed link.

for example https://invidious.snopyta.org/embed/Grv1RJkdyqI

ljoonal:
The root div of the player is not `#player-container` in invidious embeds but just `#player`. The `addHotkeyListener` function failing seems to be the root cause of all sponsor blocking failing on invidious embeds.

Fixes #645


ajayyy:
Thanks!

Issue #213, opened 2019-12-29
SponsorBlock not working on TV-friendly youtube.com/tv page


Leseratte10:
Youtube provides a TV-friendly web page at https://youtube.com/tv that looks like one of these youtube apps on a smart TV, which also allows you to "send" a youtube video from your phone to that page to watch it on a bigger screen. 

SponsorBlock currently doesn't work on that TV-optimized page, maybe support for that could be added? Mouse navigation seems to work a bit weird on that page at least in my browser (because it's optimized for TVs) so I don't know how easy it'd be to add the sponsor spot report buttons, but the playback feature (skipping spots that are already reported) should work similar to the normal YouTube page.

OfficialNoob:
Going to youtube.com/tv seems to get redirected.
The user agent "AppleTV/tvOS/9.1.1" seems to fix this :D

Leseratte10:
Oh, right, sorry, forgot to add that. Youtube seems to block normal clients from that page. This chrome addon ( https://chrome.google.com/webstore/detail/youtube-for-tv/gmmbpchnelmlmndfnckechknbohhjpge ) fixes the user agent just for this youtube page so you don't manually need to change it.

ajayyy:
Ah, I was very confused

ajayyy:
Is this related to that "smart YouTube" repo?

Leseratte10:
I don't know? Isn't that just basically a browser apk that automatically goes to youtube.com/tv? 
I just manually visit that youtube page on my computer (with that chrome addon to spoof my user agent).

OfficialNoob:
https://github.com/ajayyy/SponsorBlock/pull/214

ajayyy:
Support has been added. This issue will be left open for the preview bar support.

OfficialNoob:
Will there be buttons on the player?

ajayyy:
I don't think that is really necessary since this would be run off of a TV.

ajayyy:
Seems to be broken

itskenny0:
@ajayyy Works just fine for me as of this moment, but the UI overlays the skip interface so its only halfway usable.

nimaid:
@ajayyy This is currently not working for me.

OS: Windows 10
Browser: Firefox 80.0.1 (64-bit)
[...]

PR #701, opened 2021-03-26
Partially fix YouTube TV support

manualmanul:
This partially resolves issue #213 by using a different selector in addHotkeyListener() when viewing YouTube's TV interface at youtube.com/tv. Before this fix addHotkeyListener() would return an exception and lead SponsorBlock to assume that there are no segments in the video.

I couldn't get the SponsorBlock buttons and progress bar to work unfortunately, as YouTube TV's interface is completely different from standard YouTube.

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -1530,6 +1530,7 @@ function getSegmentsMessage(sponsorTimes: SponsorTime[]): string {
 function addHotkeyListener(): boolean {
     let videoRoot = document.getElementById("movie_player") as HTMLDivElement;
     if (onInvidious) videoRoot = (document.getElementById("player-container") ?? document.getElementById("player")) as HTMLDivElement;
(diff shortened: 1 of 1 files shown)


ajayyy:
Nice, we should probably just add a check not to fail if the element isn't found

src/content.ts
```diff
@@ -1530,6 +1530,7 @@ function getSegmentsMessage(sponsorTimes: SponsorTime[]): string {
 function addHotkeyListener(): boolean {
     let videoRoot = document.getElementById("movie_player") as HTMLDivElement;
     if (onInvidious) videoRoot = (document.getElementById("player-container") ?? document.getElementById("player")) as HTMLDivElement;
+    if (video.baseURI.startsWith("https://www.youtube.com/tv#/")) videoRoot = document.querySelector("ytlr-watch-page") as HTMLDivElement;
 
     if (!videoRootsWithEventListeners.includes(videoRoot)) {
```
ajayyy, line 1535:
```suggestion
    if (videoRoot && !videoRootsWithEventListeners.includes(videoRoot)) {
```

PR #747, opened 2021-05-14
fix wrong saved time rounding

### LINKED ISSUE #678: Time saved rounding error
59.95 minutes in time saved get rounded up to 60.0 by using toFixed(1)

https://github.com/ajayyy/SponsorBlock/blob/master/src/popup.ts#L720

https://discord.com/channels/603643120093233162/603643299961503761/820250746746175499

wilkmaciej:

Closes #678 
Also fixed small typo

### DIFF
--- a/src/popup.ts
+++ b/src/popup.ts
@@ -726,10 +726,11 @@ async function runThePopup(messageListener?: MessageListener): Promise<void> {
      * @param {float} seconds 
      * @returns {string}
      */
-    function getFormattedHours(minues) {
-        const hours = Math.floor(minues / 60);
-        return (hours > 0 ? hours + "h " : "") + (minues % 60).toFixed(1);
-    }
+	function getFormattedHours(minutes) {
+		minutes = Math.round(minutes * 10) / 10
+		const hours = Math.floor(minutes / 60);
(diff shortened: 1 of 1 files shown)


src/popup.ts
```diff
@@ -726,10 +726,11 @@ async function runThePopup(messageListener?: MessageListener): Promise<void> {
      * @param {float} seconds 
      * @returns {string}
      */
-    function getFormattedHours(minues) {
-        const hours = Math.floor(minues / 60);
-        return (hours > 0 ? hours + "h " : "") + (minues % 60).toFixed(1);
-    }
+	function getFormattedHours(minutes) {
+		minutes = Math.round(minutes * 10) / 10
```
ajayyy, line 730: I think this should be floor

wilkmaciej: Not necessarily. If there would be floor saved time will be less accurate. 
For example if saved time would be 119.9 with floor it will result in 1h 59min but with round it will be 2h 0min

wilkmaciej: It's really up to you how whould you like to handle this.

ajayyy: Ah yea, that works. I forgot this is happening before it converts to hours

PR #859, opened 2021-07-27
Make dark scroll bar in Chromium-based browsers

rakleed:

***

### DIFF
--- a/public/content.css
+++ b/public/content.css
@@ -1,3 +1,7 @@
+html {
+    color-scheme: dark;
+}
+
 #previewbar {
     overflow: visible;
     padding: 0;
--- a/public/content.css
+++ b/public/content.css
@@ -1,7 +1,3 @@
-html {
-    color-scheme: dark;
-}
-
 #previewbar {
     overflow: visible;
     padding: 0;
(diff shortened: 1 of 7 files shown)


public/popup.css
```diff
@@ -6,6 +6,10 @@
   --sb-green-bg: #077B27;
 }
 
+html {
+  color-scheme: dark;
```
ajayyy, line 10: I think this is going to make the YouTube page have dark scrollbars too

rakleed: If you add the property `html {color-scheme: dark; }` to content.css or popup.css, then when loading YouTube you can see a dark scroll bar for a split second, but then YT changes it to its own custom scrollbar, and I couldn't find the use of the standard scroll bar anywhere on YT.

public/content.css
```diff
@@ -1,7 +1,3 @@
-html {
-    color-scheme: dark;
```
ajayyy: Was there a reason for this one?

public/popup.css
ajayyy: Ah, that makes sense.

public/content.css
rakleed: Does this file change the styles for the elements that are embedded in the YT player? If so, then, indeed, this rule does not seem to make sense, since I have not seen a scroll bar anywhere in them.

ajayyy:
Thanks!

PR #887, opened 2021-08-14
Fix for an "enable category" popup

maximmax42:

***

If you choose any category when editing a segment, and then choose the "Choose a category" one, it triggers the enable category message with an empty category name. This is a fix for that.
Tested, works.


ajayyy:
good catch

PR #923, opened 2021-09-01
Mute segments

ajayyy:


2called-chaos:
Can somebody tell me what the point of this is? Just found my first mute segment and don't understand the point. I see that there is an option to disable those segments but it gets rid of them completely instead of converting them to skip.

ajayyy:
When visual content is important

Issue #953, opened 2021-09-19
Highlight button picture should stay for the entire video, not only the first few seconds.


FlorianZahn:
It would be nice if the highlight button would stay in the player. The text doesn't need to stay. I feel like at least some people will start watching a video and notice that they want to skip to the highlight because it takes too long. At this point the skip button might have already disappeared and they need to do it manually.

PR #958, opened 2021-09-20
Highlight Button now persists; Implements issue #953

FlorianZahn:

***

This makes Highlight button have the following behavior:
On timeout and skipping: Remove innertext, so that only the symbol remains. Chapters are listed after (they did disappear. I misremembered. Sorry).

I didn't change SkipButtonControlBar.disable(), because i saw it was called somewhere else.
I created a new function and replaced the disable() callse with it.

Reasoning (copied from last time):
This is great for when the highlight-button disappears before you could make the decision if you want to skip. Also you can easily skip to the highligh multiple times, which is pretty reasonable, because it is a highlight. The text disappearing only after a timeout makes sure that new users still know what the button does.

Chrome:
Click: https://www.dropbox.com/s/luvkime4squkv43/ChromeClick.mp4?dl=0
Timeout: https://www.dropbox.com/s/4ll5vbgp2d63140/ChromeTimeout.mp4?dl=0
Firefox:
Click: https://www.dropbox.com/s/sz0v3zngxazera9/FirefoxClick.mp4?dl=0
Timeout: https://www.dropbox.com/s/11xt68qyv3qfpkc/FirefoxTimeout.mp4?dl=0

The issue in question: https://github.com/ajayyy/SponsorBlock/issues/953


FlorianZahn:
I'm not sure if you like the looks of it, because it sits inbetween the time and the chapertText, but putting it after the chapterText pushed the Highlight button all the way to the right, because the chapterText thing is sooo long

FlorianZahn:
You can press "Enter" mutiple times and it will skip each time

ajayyy:
TODO: Add option to hide this button like other buttons

ajayyy:
Thanks!

PR #960, opened 2021-09-21
add start button to time edit prompt

### LINKED ISSUE #880: add start button to submit segment
I'd like to add a start button to submit segment selection (similar to (End) button that's already there). It will be useful for submitting intros on channels- I usually start segment and then have to use my mouse and type zeros in the window. I know it's quite a simple feature so let me know if there's a reason this button hasn't been added there yet. I have implementation ready if it turns out to be a good idea. Here's a screenshot of what I have in mind:
image
the button after pressing sets start segment to 0:00.000

mchangrh:

***

Add (start) button to edit time prompt & set the width to 350px

closes #880 

Tested on Chrome, scales nicely but around 1500-1700px, `(end)` protrudes to the right of the popup


public/_locales/en/messages.json
```diff
@@ -646,6 +646,9 @@
     "bracketNow": {
         "message": "(Now)"
     },
+    "bracketStart": {
+        "message": "(Start)"
+    },
```
ajayyy, line 651: I know we have other bracket ones, but I think it would be best to add the brackes in the code instead of in the localisation

src/components/SponsorTimeEditComponent.tsx
```diff
@@ -71,7 +71,8 @@ class SponsorTimeEditComponent extends React.Component<SponsorTimeEditProps, Spo
 
     render(): React.ReactElement {
         const style: React.CSSProperties = {
-            textAlign: "center"
+            textAlign: "center",
+            width: "350px"
```
ajayyy, line 75: Why limit the size?

public/_locales/en/messages.json
mchangrh: are there localizations for "start"?

src/components/SponsorTimeEditComponent.tsx
mchangrh: fixed, now using minWidth

public/_locales/en/messages.json
mchangrh: made key for "start", added brackets in code

eladkarako:
v4.7.1 I can't find it..

Issue #538, opened 2020-11-08
[RFC?] Visual Segments: Blocking out social media icons and other intrusive visual content


LoganDark:
SponsorBlock is invaluable when a portion of a video is entirely unwanted. However, some creators prefer to sprinkle subscribe buttons and social media icons throughout their video, during otherwise unrelated speech or visual activity.

Trinkets like those prefer to collect in the most inconvenient places, in the middle of useful or engaging information. Viewers generally still want that information, so skipping the section completely would not be very helpful. [Example of a video utilizing trinkets](https://www.youtube.com/watch?v=nQO_cRVpTS4): the server advertisement is a dedicated section that can be cut out with a self-promotion segment, which I have already submitted. However, the twitter trinket, which is followed by an instagram trinket, occurs in the middle of speech/gameplay that should not just be cut out.

I would like to propose a new feature for SponsorBlock: visual segments. Visual segments are not to be skipped like other segments are, but they represent a time window for a particular _trinket_ - defined as a piece of intrusive visual content, like a subscribe button or social media icon, present during an otherwise engaging or important part of the video.

Each visual segment would contain exactly one _obstruction_, which would be overlaid on top of the video content for the duration of the segment. Of course, it's not nearly as seamless as completely skipping an obnoxious section, and it's pretty obvious that there's something beneath the obstruction - but viewers won't have to see what's behind it if they don't want to, _and_ they will still not miss out on an important section of the video.

Obstructions are solid shapes (consisting of a bounding rectangle and the name of the primitive to draw, such as 'box' or 'circle'). The submitter of the obstruction can define a solid color, which defaults to black, but that's it. Obstruction bounds would be defined using percentages.

Viewers can enable or disable obstructions just like annotations. Viewers can also, in the extension settings, select if they would like to overlay the defined obstruction color, or _blur_ obstructed parts instead (achievable with backdrop-filter).

The only obstruction shapes to start would be `box` (square/rectangle) and `circle` (circle/ellipse). There's no need for fancy shapes like stars or arbitrary polygons or even rounded rectangles.

To define a visual segment, you would first select its type as 'visual segment', then you are given a basic plotting UI to define one (and only one) obstruction. It's probably up for debate how to plot multiple overlapping visual segments at once; you'd likely only do one at a time by selecting which one the plotter is using. It _sounds_ like it would be rare to see multiple social media icons on different parts of the screen at the same time... but it's actually not so rare to see, for example, one icon in each corner, during outros and such.

[...]

PR #971, opened 2021-09-26
Add start of visual segment editing

ajayyy:
Fix https://github.com/ajayyy/SponsorBlock/issues/538

***

Stored as SVGs with animations

While editing, stored as simpler types that are then converted to one svg per segment.

To play at different speeds, the svg must be modified to have all times multipled by the playback speed.

- [ ]  Implement editor with custom variables (bounds, etc.)
- [ ]  add method to convert these to an svg using svg.js
- [ ]  add custom display code for starting/stopping animations and playing at different speeds
- [ ]  add a filter on the server to only allow whitelisted xml tags


LoganDark:
This seems like a fine implementation. Clients (such as the extensions) could support whatever SVG things they want to submit to the server. Circles, rectangles, gradients etc. all fair game. Of course the official extension would likely only support basic shapes, but this looks very nice and future-proof to me.

erkinalp:
Why did you close the server side ticket? Server change is definitely required to accomodate this remedy type.

PR #985, opened 2021-10-03
Scrolling over the edit-segment-time component will change them.

FlorianZahn:
- [ ] I agree to license my contribution under LGPL-3.0 **or** my contribution is from another project with a license compatible with LGPL-3.0

***

https://streamable.com/efs7e3

### DIFF
--- a/src/components/SponsorTimeEditComponent.tsx
+++ b/src/components/SponsorTimeEditComponent.tsx
@@ -132,6 +132,7 @@ class SponsorTimeEditComponent extends React.Component<SponsorTimeEditProps, Spo
                                         timeAsNumber = 0;
                                     }
                                     sponsorTimeEdits[0] = utils.getFormattedTime(timeAsNumber, true);
+                                    if (getCategoryActionType(sponsorTime.category) === CategoryActionType.POI) sponsorTimeEdits[1] = sponsorTimeEdits[0];
                                     this.setState({sponsorTimeEdits});
                                     this.saveEditTimes();
                                 }
--- a/src/components/SponsorTimeEditComponent.tsx
+++ b/src/components/SponsorTimeEditComponent.tsx
@@ -56,6 +56,11 @@ class SponsorTimeEditComponent extends React.Component<SponsorTimeEditProps, Spo
             event.stopPropagation();
         });
 
+        // Prevent scrolling while changing times
+        document.getElementById("sponsorTimesContainer" + this.idSuffix).addEventListener('wheel', function (event) {
+            event.preventDefault();
+        }, {passive: false});
+
         // Add as a config listener
         if (!this.configUpdateListener) {
             this.configUpdateListener = () => this.configUpdate();
@@ -86,7 +91,6 @@ class SponsorTimeEditComponent extends React.Component<SponsorTimeEditProps, Spo
                 node.style.setProperty("text-shadow", "none", "important");
             }
         };
-
         // Create time display
         let timeDisplay: JSX.Element;
         const sponsorTime = this.props.contentContainer().sponsorTimesSubmitting[this.props.index];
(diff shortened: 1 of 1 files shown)


src/components/SponsorTimeEditComponent.tsx
```diff
@@ -114,6 +117,25 @@ class SponsorTimeEditComponent extends React.Component<SponsorTimeEditProps, Spo
...
+                                        timeAsNumber += step;
+                                    }else if (timeAsNumber >= step){
+                                        timeAsNumber -= step;
+                                    } else {
+                                        timeAsNumber = 0;
+                                    }
+                                    sponsorTimeEdits[0] = utils.getFormattedTime(timeAsNumber, true);
+                                    if (getCategoryActionType(sponsorTime.category) === CategoryActionType.POI) sponsorTimeEdits[1] = sponsorTimeEdits[0];
+                                    this.setState({sponsorTimeEdits});
+                                    this.saveEditTimes();
```
ajayyy, line 137: Can you extract this to a function taking in an `index`?

src/components/SponsorTimeEditComponent.tsx
```diff
@@ -240,6 +282,18 @@ class SponsorTimeEditComponent extends React.Component<SponsorTimeEditProps, Spo
...
+        // Returns true if the wheel scrolls up and false if the wheel scrolls down
+        console.log("wheelUpOrDown")
+        if (e.deltaY < 0)
+        {
+            return true;
+        }
+        else if (e.deltaY > 0)
+        {
+            return false;
+        }
```
ajayyy, line 295:
```suggestion
        return e.deltaY < 0;
```

src/components/SponsorTimeEditComponent.tsx
```diff
@@ -240,6 +282,18 @@ class SponsorTimeEditComponent extends React.Component<SponsorTimeEditProps, Spo
         );
     }
 
+    wheelUpOrDown(e: React.WheelEvent): boolean{
+        // Returns true if the wheel scrolls up and false if the wheel scrolls down
```
ajayyy, line 286: Instead of the comment, just change the name to `isWheelScrollingUp`

src/components/SponsorTimeEditComponent.tsx
FlorianZahn: I'm not sure if `deltaY` can be = 0

FlorianZahn: [It seems like it can be 0](https://www.w3schools.com/jsref/event_wheel_deltay.asp)

> otherwise 0

I dont know how to handle this. Could the whole thing crash if it is 0?

src/components/SponsorTimeEditComponent.tsx
```diff
@@ -240,6 +247,25 @@ class SponsorTimeEditComponent extends React.Component<SponsorTimeEditProps, Spo
...
+        const sponsorTimeEdits = this.state.sponsorTimeEdits;
+        let timeAsNumber = utils.getFormattedTimeToSeconds(this.state.sponsorTimeEdits[index]);
+        if (timeAsNumber !== null && e.deltaY != 0) {
+            if (e.deltaY < 0){
+                timeAsNumber += step;
+            }else if (timeAsNumber >= step){
+                timeAsNumber -= step;
+            } else {
+                timeAsNumber = 0;
+            }
```
ajayyy, line 261: if it is zero, ut just doesn't need to be changed, it doesn't have to be reset to zero.

FlorianZahn: This more for if the time left is 0.003 seconds. It would stay without the  the else.

FlorianZahn: `timeAsNumber >= step` isn't needed actually, because `utils.getFormattedTimeToSeconds` only return `>=0`, but if this were to be changed, it wouldnt be nice if this piece of code broke because of it.

FlorianZahn:
Maybe also add the ctrl key to change some bigger amoutn, like 0.1, 0.5 or 1 second.

FlorianZahn:
The message still needs to be translated. I'm not sure how to add it, so that it appears on crowdin, though.

PR #1002, opened 2021-10-13
Added margin-bottom for #sponsorblockPopup

maximmax42:
This margin is to separate popup window from playlist/related videos list.

***

### DIFF
--- a/public/popup.css
+++ b/public/popup.css
@@ -46,6 +46,7 @@
   width: 330px;
   padding: 22px;
   text-align: center;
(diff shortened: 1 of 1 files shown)


maximmax42:
Oh drat. Didn't realize it would be applied to that as well. I'll come back to you on that when I will be able to check.

ajayyy:
https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties#what_happens_with_invalid_variables

PR #1029, opened 2021-10-29
Add more detail to building section of readme

opl-:

Included links to relevant browser documentation, added an explanation for the mysterious `CompileConfig` errors, and made the command to build the development version the more prominent variant (as it's arguably more useful in development).

Preview [here](https://github.com/opl-/SponsorBlock/tree/docs/readme-building#building).

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,21 +60,24 @@ You can read the API docs [here](https://wiki.sponsor.ajay.app/index.php/API_Doc
 
 # Building
 
-You must have Node.js 16 installed.
+You must have [Node.js 16](https://nodejs.org/) and npm installed.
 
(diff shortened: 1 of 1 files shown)


opl-:
I removed that bit because it was rather confusing as it was. How should an average person adjust the configuration? Why should they do it?

Realistically, I feel like for most people who just want to contribute to the mainline extension telling them to "adjust the config" ends up causing more confusion than helping.

If you think it's important, it might be worth either creating a separate readme section, or linking to an appropriate wiki page which would explain why it'd be useful to do it.

ajayyy:
changing the server url to localhost for easier testing

PR #1071, opened 2021-11-30
Auto-populate invidious list

### LINKED ISSUE #948: Auto-fetch `invidious` instances from instance list
### Feature request
If `Support 3rd Party YouTube-Sites`is toggled on, periodically automatically scrape invidious instances from the [instance list](https://invidio.us/).

This will eliminate the need to manually configure invidious instances most of the time.
However I believe these domains should also be blacklist-able, since it may not always be wanted to automatically block video segments on all of these sites at the same time.

I briefly went through the source code of the instance list's [`main.js`](https://git.sr.ht/~cadence/invidious-redirect/tree/master/item/src/main.js) and spotted following [`instances.json`](https://api.invidious.io/instances.json).

That `.json` file is from where it loads the currently up and running instances, it should also be usable to implement this feature in SponsorBlock. :slightly_smiling_face:

mchangrh:

***

closes #948


ajayyy:
This will need to somehow "lock" the value of `invidiousInstances` for users with Invidious enabled, as they won't have permissions for the new domains. If a default value is never adjusted, then it is never set in the storage.

One way would be to add a check in the migrate function in `config.ts` and just set `invidiousInstances` to itself.

mchangrh:
I'm not sure I understand. If a user have overridden the `invidiousinstances` list, then the default wouldn't apply and "reset" should apply the new detauls

wouldn't a migrate only be needed for users with only `["invidious.snopyta.org"]`?

mchangrh:
Ah, I see now.

Correct me if I'm wrong but this should fix it
- > migrate function in config.ts and just set invidiousInstances to itself.
  - don't migrate if invidious enabled
- reset takes from default

ajayyy:
Only migrate if invidious is **enabled**, as that means permissions have been granted, and we can't adjust which domains are invidious instances.

src/config.ts
```diff
@@ -433,6 +433,11 @@ function migrateOldFormats(config: SBConfig) {
     if (config["previousVideoID"] !== undefined) {
         chrome.storage.sync.remove("previousVideoID");
     }
+
+    // populate invidiousInstances with new instances if 3p support is **DISABLED**
+    if (!config["supportInvidious"]) {
+        config["invidiousInstances"] = invidiousList
+    }
```
ajayyy, line 440:
```suggestion
    if (!config["supportInvidious"] && config["invidiousInstances"] !== invidiousList) {
        config["invidiousInstances"] = invidiousList;
    }
```

maybe this will save a call to chrome.storage.set each reload? Not sure if the equality will work though

mchangrh: doesn't work, it modifies it every time

mchangrh: maybe just a simple `.length` check?

ajayyy: yea, probably good enough

mchangrh: `.length` works like a charm

PR #1093, opened 2021-12-13
Settings rework

### LINKED ISSUE #615: Improve server address settings clarity
Instead of 
current
I would recommend doing it like
new

Benefits:

- always clear, which option is activated (e.g. currently beta toggle doesn't change the URL)
- main URL can be changed by the extension at any time (if so needed) without changing users' config
- no reset button needed because the user can just choose another radio button

### LINKED ISSUE #616: Change default shortcuts based on locale
<kbd>Shift</kbd>+<kbd>,</kbd> is already used in YouTube for lowering the playback speed. Every time i want to slow down the speed a new segment is marked or ended in Sponsor Block.
Please choose a new shortcut for that function, so it doesn't overlap with builtin functionality anymore :)

### LINKED ISSUE #757: Button to remove shortcut and set it to "unset"
I think it would be very useful if one could click on a button or press `ESC` (or something similar) in the _Set key for ..._ menus  to reset the shortcut and set it to `NULL`/`UNSET`

AronHK:
- I agree to license my contribution under LGPL-3.0 **or** my contribution is from another project with a license compatible with LGPL-3.0

***

- added tabs to Settings and grouped options into categories
  - link to specific tabs by using their ID as anchor, like: _options.html#keybinds_
- options that depend on others _(Auto-hide Info Button ->  Show Info Button On YouTube Player)_ get hidden/shown without a refresh
- added more dependencies
  -  Show Delete Button On YouTube Player, Show Upload Button -> Show Buttons On YouTube Player
  - Enable Skip Count Tracking In Private/Incognito tabs -> Enable Skip Count Tracking
  - SponsorBlock Server Address -> Enable Beta Testing Server (resolves #615)
  - Add 3rd-Party Client Instance -> Support 3rd Party YouTube-Sites
- added missing labels to some options
- changed things to be HTML5 compliant
- added support for light theme (help page too, so the embed doesn't look out of place)
- show version number in settings
- redesigned keybind settings (resolves #757)
- added support for using ctrl, alt and shift in shortcuts (resolves #616)
- option to save/load settings json to file


AronHK:
I can remove the light theme CSS if you really think it would be a pain to maintain. But it's basically choosing two colors instead of one in the rare case you need it, so I don't think it's too bad.

src/options.ts
```diff
@@ -30,39 +30,41 @@ async function init() {
     await utils.wait(() => Config.config !== null);
 
     if (!showDonationLink()) {
-        document.getElementById("sbDonate").style.visibility = "hidden";
+        document.getElementById("sbDonate").style.display = "none";
```
ajayyy, line 33: This was to prevent layout shifts

AronHK: I'm not sure what layout shifts you mean. That the "created by" text recenters itself? I thought that would be desirable. Especially because I moved things around a bit and it being off-center is more noticeable.

public/options/options.html
```diff
@@ -13,608 +14,478 @@
...
+				<div option-type="toggle" data-sync="fullVideoSegments">
+					<div class="switch-container">
+						<label class="switch">
+							<input id="fullVideoSegments" type="checkbox" checked>
+							<span class="slider round"></span>
+						</label>
+						<label class="switch-label" for="fullVideoSegments">
+							__MSG_fullVideoSegments__
+						</label>
 					</div>
```
AronHK, line 109: I put it under mute, but should this be in Interface instead? It doesn't affect anything but whether it shows the icon, right?

ajayyy: I feel that this is like enabling or disabling a skip option since this is another way to label things.

AronHK: The three relevant tabs I created are Categories, Behavior and Interface.
I only put the actual categories into Categories. (I'm assuming that's what you mean by skip option?)
I put "Auto skip all segments when there is a non-music segment" and "Allow segments that mute audio instead of skip" into Behavior. This option doesn't technically change skipping behavior, but I suppose it's still a functionality, so it makes sense for it to be there.
Or would you want these 3 options to be on the same tab as the categories?

ajayyy: Just took a look, could those three options all be with the categories at the bottom?

AronHK: okay

ajayyy:
- Coulda max width be set to not stretch the to the entire page like the old options?
- I don't know about the default white theme... Could it maybe be dark unless a specific toggle has been triggered?
- The version loading in seems to cause a layout shift on Chrome. Maybe that could be switched to absolute positioning or a set height.

ajayyy:
I think the number of categories are a little overwhelming. Maybe "Privacy" and "Advanced" can be combined into "Miscellaneous". And maybe "Categories" can be merged into "Behavior".

Current:

image

Change:

image

Also, can't figure out how to do it, but if it is possible with flex, maybe have them not stretch like below:

image

AronHK:
>I don't know about the default white theme... Could it maybe be dark unless a specific toggle has been triggered?

It's not default, it follows your browser theme. Looks like you are using light mode 😄
In Firefox you can change your theme, Chrome seems to always follow the OS theme.
There could be a toggle to change the theme, but I would still follow the browser by default.

>I think the number of categories are a little overwhelming.

I made Privacy separate because some people will instantly be looking for it. But it's pretty empty, so it makes sense to merge it.
Behavior+Categories would be a pretty long list again, which is what I was trying to solve. Though with those 3 options already moved the difference wouldn't be huge. I liked the category table being completely separate, but it's your call.

ajayyy:
Chrome doesn't follow theme properly everywhere, such as on Linux. And with Chrome, switching to the "dark" theme (for the tabs and other bars) doesn't actually switch the it officially to "dark", so it still defaults to light theme.

AronHK:
Didn't even think of that..
kép
But of course this doesn't stop them from sending _prefers-color-scheme: light_...

Maybe on Chrome+Linux we could manually set _prefers-color-scheme: dark_?
Are Windows 8 and older Mac versions worth worrying about?

ajayyy:
Hmm, I see the issue now as it's not really possible to support a light/dark theme without loading it in JS that doesn't use `prefers-color-scheme`. And if it is in JS, then you'll get a flash of white. My biased opinion says to just not have a light mode at all, but I'll post a question in discord.

src/options.ts
AronHK: The best solution I could come up with that wasn't off-center or pushing anything around was putting it into the corner:
kép

AronHK:
@ajayyy I think I resolved everything you mentioned

ajayyy (approved):
Thanks a lot, it looks really nice!

AronHK (commented):
I already limited the width by setting the right padding of #options to 15%. With the sidebar, that resulted in about the same width as in the old settings page. If you prefer the centered look (and using max-width) you should probably remove left-right paddings from #options. (Don't forget the ones in the media queries.)

I also think 60% with the sidebar is too small. I would personally go with 80% by default and only switch to 60% in `@media only screen and (max-height: 770px), only screen and (max-width: 1200px)`

AronHK:
This is what it looked like:
kép
The actual content is about the same width as on the old page:
kép
(Looking at the screenshots it's slightly wider, but not a lot.)
The current version looks like this:
kép

Did you only see the version with the top menubar? That should only appear if height is lower than 770px or width is lower than 1200px. I did forget to limit the width for that one.

ajayyy:
Ah, I didn't know about this layout as my monitor isn't wide enough for that.

ajayyy:
Are there other layouts you added? If so, could you send some screenshots?

AronHK:
Just these two. It shows the menu on the side, until it doesn't fit, then it moves to the top. Other than that I was only fiddling with the margins as the window size gets smaller.

Do you want me to make a new PR to sort out the sidebar-layout, or do you want to do it?

ajayyy:
Thanks, I can figure out the sidebar-layout issue

ajayyy:
Fixed it up here: 88f61cf292e9401e03868d434131f74bdb1a9a32

PR #1100, opened 2021-12-20
Improve keybind options

wbholland:

***
- fixes bug where user could bind two actions to the same key
- allows user to only select one keybind option at a time
- written to allow for additional keybind options in the future

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,7 +7,7 @@
     "": {
       "name": "sponsorblock",
       "version": "1.0.0",
-      "license": "GPL-3.0-only",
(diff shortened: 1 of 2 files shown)


AronHK:
I'm also modifying keybinds as part of #1093. Sorry, it wasn't stated in the pull request, only in an issue.
I changed the UI, so changing multiple keybinds at the same time isn't possible anymore.
The change for checking conflicts could probably be compatible with mine, but with some changes, since I'm not using a single key anymore.

ajayyy:
I merged @AronHK change for this instead, but thanks a lot for looking into it!

PR #1110, opened 2021-12-29
fix skipNotice duration with playbackrate != 1

### LINKED ISSUE #1067: Playback speed is not correctly taken into account for the remaining seconds of a skip notice
When the playback speed is x2 for example the remaining time is divided by 2 but the timer also goes twice as fast so it's like dividing the time by 4. And conversely for a speed of x0.25. The remaining time will be multiplied by 4 but the timer will go 4 times slower so it's as if it was 16 times longer.
It is possible that it is a problem on my side but I tested with firefox and chrome and I have the same results.

mchangrh:

closes #1067 

***


ajayyy:
I think it's better removing it at the other place 2db19711900e0a87ee8cb97df2d8645375a24413

PR #1113, opened 2021-12-30
fix leak that was killing performance over long periods

CyberPhoenix90:

***

Hey I noticed that over several hours of use the CPU usage of youtube music would get worse and worse. After a quick check I found that the set interval here is piling up if the condition never becomes true. I was having hundreds of setIntervals firing every second.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -26,7 +26,10 @@ export default class Utils {
     /** Function that can be used to wait for a condition before returning. */
     async wait<T>(condition: () => T | false, timeout = 5000, check = 100): Promise<T> {
         return await new Promise((resolve, reject) => {
-            setTimeout(() => reject("TIMEOUT"), timeout);
+            setTimeout(() => {
+                clearInterval(interval);
+                reject("TIMEOUT")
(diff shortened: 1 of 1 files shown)


src/utils.ts
```diff
@@ -26,7 +26,10 @@ export default class Utils {
     /** Function that can be used to wait for a condition before returning. */
     async wait<T>(condition: () => T | false, timeout = 5000, check = 100): Promise<T> {
         return await new Promise((resolve, reject) => {
-            setTimeout(() => reject("TIMEOUT"), timeout);
+            setTimeout(() => {
+                clearInterval(interval);
```
ajayyy, line 30: How is this compiling. Shouldn't this be moved to after defining the interval?

ajayyy (commented):
Good catch, thanks!

src/utils.ts
CyberPhoenix90: I can move it if you wish but the reason this is valid is because the use of interval here occurs in a closure. Closures can access variables that are defined after the fact.
image
As demonstrated here

Issue #1153, opened 2022-01-26
Mobile YouTube whitelist doesn't work


yoasif:
Steps to reproduce:

1. Follow instructions on https://blog.mozilla.org/addons/2020/09/29/expanded-extension-support-in-firefox-for-android-nightly/ to install Sponsorblock 
2. Go to https://m.youtube.com/watch?v=xf_ReluPZf4
2. Go to add-ons, Sponsorblock
3. Tap the "whitelist channel" link

What happens:

I see an error message saying that the channel id is not loaded yet. 

Expected result:

The channel is whitelisted. 

---

Additionally, if you whitelist the channel in the desktop view and switch back to the mobile view, the whitelist does not work in that view (it works in the desktop view).

mchangrh:
Have you tried `Force Channel Check Before Skipping` ?

mobile YT seems to use different attributes for channel name & logo

mchangrh:
fixed in #1155

Issue #1053, opened 2021-11-23
Media buttons stop working after skipping a segment with audio notification on skip


Danstiv:
### Hardware and software
Some bluetooth headphones called bt12.
Logitech keyboard k 270.
I think any device with multimedia buttons will do.
Windows 10 21H1 (64-bit) build 19043.1052
Google Chrome 96.0.4664.45
SponsorBlock (installed from chrome webstore) 3.6.1.
### Steps to reproduce
1. Connect your device with multimedia buttons to your computer.
2. Open the video for which the segments were sent (for example [this](https://www.youtube.com/watch?v=2USiC66uIMQ)).
3. Make sure the media buttons are working (just check play / pause).
4. Continue watching to segment.
5. After skipping (auto skip must be enabled) check the multimedia buttons again.
### Actual behavior
Media buttons no longer work
### Expected behavior
Multimedia buttons continue to work after skipping a segment.
### Notes
As mentioned above, I used bluetooth headphones as a device with multimedia keys.
Also, my logitech k270 keyboard has a play / pause button that works before skipping and stops working after skipping a segment.

mchangrh:
Could not reproduce 


multimedia keys work on wired keyboard/ media remapped keys (Corsair K70/ Logi G502) and through bluetooth (Aukey BR-C1) albeit delayed

W10 21H1 19043.1387"
Chrome 96.0.4664.45
SB 3.6.2

Danstiv:
I found a setting that affects this.
I have the audioNotificationOnSkip setting enabled.
It really works, when a segment is skipped, a sound signal is played.
And, apparently, at the same time something happens with the audio focus, because of which the multimedia buttons stop working.
Try to reproduce with the sound notification turned on.

mchangrh:
I can recreate it consistently with audioNotificationOnSkip

madushans:
I can reproduce this (or something very similar) on Edge.

Edge 97.0.1072.62
Windows 11 21H2 22000 434
Pay/Pause button mapped to a custom mouse button
SponsorBlock configured to skip sponsor and non-music blocks

When a segment is auto skipped at the beginning, Edge does not respond to the pay/pause button.

When SponsorBlock is configured to only show the block in seekbar and not auto skip.
Skipping blocks manually does not cause this issue it seems.

Disabling the audio notification on skip as @Danstiv mentioned fixes the issue.

mchangrh:
fixed in #1155

PR #1155, opened 2022-01-26
Misc fixes

mchangrh:

***

I just got really distracted and now I have a messy working codebase

- [x] Replace String.substr with String.slice
- [x] Fix misc eslint errors
- [x] Fix mobile ChannelID lookup #1153 
- [x] Fix taking audio focus #1053 
  - set mediaSession metadata to null and then set to old metadata


src/content.ts
```diff
@@ -971,8 +967,7 @@ function getYouTubeVideoIDFromURL(url: string): string | boolean {
         return id.length == 11 ? id : false;
     } else if (urlObject.pathname.startsWith("/embed/") || urlObject.pathname.startsWith("/shorts/")) {
         try {
-            const id = urlObject.pathname.split("/")[2];
-            if (id && id.length >= 11) return id.substr(0, 11);
+            return urlObject.pathname.split("/")[2].slice(0, 11);
```
ajayyy, line 970: Why this change?

mchangrh: slice won't return a different value if it's already under the length, adding the check for id and id.length isn't necessary

mchangrh: ah I see what you mean

mchangrh: reverted

Issue #1159, opened 2022-02-01
never skip for youtube.com/clip/


mchangrh:
clips should never contain skip segments since it was linked to that specific time
it also seems to break the UI https://discord.com/channels/603643120093233162/603643256714297374/938019615735431198, might be because it highlights the full video

HugoHead:
youtube.com/clip/ redirects to the a channel called "clips" 

what are you referring to?

mchangrh:
> youtube.com/clip/ redirects to the a channel called "clips"
> 
> what are you referring to?

Example: https://www.youtube.com/clip/UgkxeLPGsmKnMdm46DGml_4wd2dpAOIFJ65l

mchangrh:
.take

mchangrh:
YouTube doesn't send any events for going from clip to full video, the videoDuration is actually the length of the full video & the URL is likely random, it doesn't render any base64 results. videoID change is not triggered

if this ever changes, in theory it is possible to get the videoID from the ytCommand object created

```
ytWindow.ytCommand?.watchEndpoint?.videoId
```

PR #1184, opened 2022-02-09
ytplayer.config fallback

mchangrh:
taken from #1159

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -961,6 +961,15 @@ function getYouTubeVideoIDFromDocument(document: Document, hideIcon = true): str
 }
 
 function getYouTubeVideoIDFromURL(url: string): string | boolean {
+    // get ID from ytPlayer object
+    interface ytPlayerWindow extends Window {
+        ytplayer: { config: { video_id: string; }}
+    }
+    // fall back to window.ytPlayer object
+    const ytWindow = (window as unknown as ytPlayerWindow);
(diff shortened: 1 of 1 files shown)


src/content.ts
```diff
@@ -999,6 +999,16 @@ function getYouTubeVideoIDFromURL(url: string): string | boolean {
             return false;
         }
     }
+
+    interface ytPlayerWindow extends Window {
```
ajayyy, line 1003: It won't get here since searchParams has v, but id.length is not 11

mchangrh: moved to top, since it should be the most reliable if it does exist

ajayyy:
Todo: test how fast this changes. Does url change first?

mchangrh:
put it in URL search since it's only applicable on the youtube site, there's also ytInitialData which might be another fallback

mchangrh:
also I think window.ytplayer might be a/b tested since i'm not seeing it in private window, that or it's only on specific pages

ajayyy:
ytInitialData only is valid for the first video, but stays in the page, so I don't think it's a good idea to rely on

mchangrh:
I misunderstood

mchangrh:
seems like ytcfg might also be a reliable source of videoID

mchangrh:
nope, window.ytplayer follows ytinitialdata in playlists

PR #1189, opened 2022-02-13
Add menu bar to popup

AlecRust:
### Current Problem

New user here - great service! Only area I was slightly confused by was the "popup" that appears when you click the extension icon. I found it wasn't immediately clear what the main actions were.

uBlock Origin for example has clear [menu bars](https://i.imgur.com/Yr9LoFM.png) that I think might help improve the structure of this area.

### Proposed Solution

This PR refactors the "popup" HTML and CSS to introduce a basic menu bar for the main actions.

I believe this PR to be production-ready, but it's only a "first pass" at improving this area of the extension. If this gets merged and you like the direction this is heading I'm happy to follow it with some more areas of improvement I see.

### Screenshots

**On YouTube page:**

Before             |  After
image  |  image

**Not on YouTube page:**

Before             |  After
image  |  image

---

### DIFF
--- a/public/popup.css
+++ b/public/popup.css
@@ -1,235 +1,274 @@
 :root {
+  --sb-main-font-family: "Source Sans Pro", sans-serif;
   --sb-main-bg-color: #222626;
-  --sb-main-fg-color: white;
-  --sb-gray-fg-color: #444848;
-  --sb-on-white-bg: black;
-  --sb-green-bg: #077B27;
+  --sb-main-fg-color: #fff;
+  --sb-grey-bg-color: #333;
+  --sb-red-bg-color: #cc1717;
 }
+
+/*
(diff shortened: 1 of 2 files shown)


public/popup.css
```diff
@@ -1,236 +1,275 @@
...
 
-@media only screen and (max-width: 600px) {
-  #sponsorBlockPopupBody {
-    width: 100%; 
-  }
+button {
+  background: transparent;
+  border: 0;
+  cursor: pointer;
+  padding: 0;
```
ajayyy, line 36: It's not possible to have generic selectors like this as this css is used on YouTube for the popup on the webpage, so the CSS can't interfere.

ajayyy:
thanks!

PR #1192, opened 2022-02-14
fix: ignore switchingVideos in Safari

### LINKED ISSUE #1142: Auto skip doesn't work in Safari when autoplaying in a playlist
If you play a video in a playlist the skip works as expected, however when YouTube moves on to the next video in the playlist it doesn't work for that video (Or any subsequent video) unless you manually seek in the video or select a new video in the playlist.

Let me know if there's any way to enable debug info or similar.

Here's an example: https://youtu.be/i9St4aTUVis?list=PL17i3ijKy9Rcoui9uZ-tSwDH2j-mbgLw7&t=716

The timestamp is set just before a sponsor segment, which will be skipped. That video then ends shortly after and the next video has a sponsor segment near the start which will not be skipped.

asportnoy:

***

This disables the `switchingVideos` check in the `startSkipScheduleCheckingForStartSponsors` function for Safari as discussed in #1142.

Closes #1142

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -879,7 +879,8 @@ function retryFetch(): void {
  * Ex. When segments are first loaded
  */
 function startSkipScheduleCheckingForStartSponsors() {
-    if (!switchingVideos && sponsorTimes) {
+	// switchingVideos is ignored in Safari due to event fire order. See #1142
+    if ((!switchingVideos || isSafari) && sponsorTimes) {
         // See if there are any starting sponsors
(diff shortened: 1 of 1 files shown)


purrloftruth:
Would love to test this fix as I've been suffering from the same issue but there is no build artifact for the Safari extension, and the project's readme only has information on building the Chrome and Firefox versions.

its0x08 (approved):

ajayyy:
Seems to work in my testing, I did have one case where it still didn't work, but I could not reproduce it

PR #1227, opened 2022-03-12
Refactor segment expanding to use <details>

AlecRust:
### Current Problem

- It's not very clear that segments can be expanded/collapsed. A hover state would help indicate each segment can "do something" when clicked.
- The expand/collapse of these time segments is done via JS, which is unnecessary considering the [`<details>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details). It also has subtle issues like being unable to collapse after voting.

### Proposed Solution

- Add hover state to time segments
- Replace JS expand/collapse with `<details>` element

Also:
- Remove unused `.popupElement` class

### Screenshots

Before             |  After
Before  |  After

***

### DIFF
--- a/public/popup.css
+++ b/public/popup.css
@@ -79,18 +79,26 @@
 }
 
 /*
- * Individual segments (<button> elements)
+ * <details> wrapper around each segment
  */
 
-.segmentTimeButton {
-  background: transparent;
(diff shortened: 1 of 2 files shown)


ajayyy:
very cool!

Issue #594, opened 2020-12-23
Keep info button available when popup is shown


Madis0:
It is bad UX to open the popup via the info button, but not being able to close it from there (and having to look for a small "close popup" link instead).

I suggest making the button show "info icon" by default and "close icon" (a cross inside of the SB outline) when the popup is open. That would make it consistent with the start/stop button.

Additionally, the in-page popup could also have a standard X on the top right corner, similar to the "submit segment" popup, so the user can use either way to close it.

AlecRust:
@Madis0 I think we can close this, the close button is now implemented. I would create a separate issue for anything else outstanding.

PR #1228, opened 2022-03-13
Fix popup width and improve close button

AlecRust:
### Current Problem

- The popup width doesn't match the YouTube sidebar width when displayed in-page.
- The "Close Popup" link is very small and easily missed. I would expect a close icon top right, similar to what is mentioned in #594.
- The "You've saved" and "You've skipped" paragraphs aren't on their own line (small regression in #1189) ([screenshot](https://i.imgur.com/VqeXMmS.png)).

### Proposed Solution

- Allow width of popup to fill container when displayed in-page.
- Move the "Close Popup" button to the top right of the popup as an icon.
- Use `display: block` instead of `display: unset` for showing "You've saved" and "You've skipped" paragraphs.

### Screenshots

Before             |  After
image  |  image

***

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -34,7 +34,7 @@ let sponsorVideoID: VideoID = null;
 const skipNotices: SkipNotice[] = [];
 let activeSkipKeybindElement: ToggleSkippable = null;
 
-// JSON video info 
+// JSON video info
 let videoInfo: VideoInfo = null;
 // The channel this video is about
 let channelIDInfo: ChannelIDInfo;
@@ -221,8 +221,8 @@ function messageListener(request: Message, sender: unknown, sendResponse: (respo
 
 /**
  * Called when the config is updated
- * 
- * @param {String} changes 
+ *
+ * @param {String} changes
  */
 function contentConfigUpdateListener(changes: StorageChangesObject) {
     for (const key in changes) {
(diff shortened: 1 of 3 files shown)


AlecRust:
Apologies for the whitespace changes in the diff applied by my editor. I recommend viewing with this enabled:
image

its0x08 (approved):

src/content.ts
```diff
@@ -1638,17 +1638,22 @@ function openInfoMenu() {
...
+            const closeButtonIcon = document.createElement("img");
+            closeButtonIcon.src = chrome.extension.getURL("icons/close.png");
+            closeButtonIcon.width = 16;
+            closeButtonIcon.height = 16;
+            closeButton.appendChild(closeButtonIcon);
+            closeButton.setAttribute("title", chrome.i18n.getMessage("closePopup"));
+            closeButton.classList.add("sbCloseButton");
             closeButton.setAttribute("align", "center");
             closeButton.addEventListener("click", closeInfoMenu);
             // Theme based color
```
ajayyy, line 1651: Probability should not change colour based on theme anymore

ajayyy:
Will have to test on ultra-wide displays, but I remember it appearing pretty ugly there

AlecRust:
My resolution is 3440×1440, seems ok. But more testing the better!

src/content.ts
AlecRust: Is this a change for this PR?

ajayyy: Yes, otherwise the close button will be white on light theme, even though it is inside the SB popup which is always dark

AlecRust: Thanks, I've removed this.

mchangrh:
LGTM on 2560x1080 Ultrawide

Theater + default modes both look fine

PR #1237, opened 2022-03-19
Improve popup alert and fix Firefox issues

### LINKED ISSUE #778: Scrollbars when you have SponsorBlock on overflow menu in Firefox.
### Description
There is scrollbars when you put SponsorBlock to Overflow menu in Firefox.

image

### Steps to Reproduce

1. Put SponsorBlock addon to overflow menu in Firefox.
2. Click to open it

### Expected Behavior

No scrollbars.

### LINKED ISSUE #1235: [UI] setUsername button clips into username
it seems like
```css
margin-right: 8px;
```
is somehow being overridden by another CSS rule, causing the pencil icon to clip into the username

image

AlecRust:
### Current Problem

- The "consider enabling force channel check" alert has wrapping text and pushes the content down quite a lot
- The alert doesn't disappear when the thing that triggered it is toggled off
- Couple minor spacing regressions in #1189
- Forced popup width causes horizontal scrollbar/misalignment in Firefox
- Lack of `max-height` can cause Firefox popup to be too large/cause scroll issues

### Proposed Solution

- Use colours instead of font size/weight to make the alert noticeable
- Hide the alert when "Remove channel from whitelist" is clicked
- Fix some spacing regressions
- Add `max-width: 100%` to fix Firefox popup width issues
- Add `max-height: 600px` (only in popup context) to fix Firefox popup height issues

This also tidies/simplifies the popup CSS a bit.

### Screenshots

Before             |  After
before  |  after

***

closes #778
closes #1235

### DIFF
--- a/public/popup.css
+++ b/public/popup.css
@@ -29,6 +29,8 @@
 
 #sponsorBlockPopupHTML {
   color-scheme: dark;
+  max-height: 600px;
+  overflow-y: auto;
 }
 
 #sponsorBlockPopupBody {
--- a/public/popup.css
+++ b/public/popup.css
@@ -475,7 +475,7 @@
   overflow: hidden;
   text-overflow: ellipsis;
   max-width: 130px;
-  margin-right: 8px;
+  margin: 0 8px 0 0;
 }
 
 /*
@@ -531,16 +531,12 @@
 }
 
 /*
- * Generic spacing classes
+ * Generic utilities
  */
 
-.u-mZ {
+ #sponsorBlockPopupBody .u-mZ {
   margin: 0 !important;
 }
-
-/*
- * Generic hide utility classes
- */
 
 #sponsorBlockPopupBody .hidden {
   display: none !important;
--- a/public/popup.css
+++ b/public/popup.css
@@ -317,7 +317,7 @@
 
 .sponsorStartHint {
   display: block;
-  margin-bottom: 12px;
+  padding: 0 10px 12px;
 }
 
 /*
@@ -379,7 +379,7 @@
 .sbYourWorkCols > div {
   display: flex;
   align-items: center;
-  width: 50%;
+  flex-basis: 50%;
   justify-content: center;
 }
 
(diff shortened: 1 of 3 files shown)


AronHK (commented):
the in-page popup also gets a scrollbar (this may be fine) and the close button appears slightly over it.
the enable/disable toggle also breaks here.
image

Nojuuu:
> the in-page popup also gets a scrollbar (this may be fine) and the close button appears slightly over it. the enable/disable toggle also breaks here. 

Maybe that in-page popup could be handled with `:not()`?

AlecRust:
Thanks, I will take a look at the "enable/disable toggle also breaks" issue.

Regarding the scrollbar appearing when rendered in-page, and close icon appearing over scrollbar, some ideas:

- Simply disable max height/scrollbar when rendered in-page
- Keep max height/scrollbar when rendered in-page, and position the close button on the left (always visible)
- Keep max height/scrollbar when rendered in-page, and keep the close button on the right but fix the position issue i.e. it would shift over if scrollbar appeared, and always be visible
- Keep max height/scrollbar when rendered in-page, and keep the close button on the right but make it only visible when your scroll position is at the top of the container

Happy with any option, if you guys have any preference?

Really the max height/scrollbar stuff should be removed from this PR and done separately as the changes here are getting a bit off track. If it's significant code to fix the above I may do that.

AronHK:
I would vote A or C.
The toggle breaking only happened for me when a scrollbar was present, so doing A may be the easiest. (I had something similar happen when I was messing around with the popup, I think it had to do with elements of the toggle being absolutely positioned.)

AlecRust:
Thanks, agreed.

I've disabled the max height/scrollbar when it's rendered in-page now, if you want to test. The max height remains just as a fix for the Firefox popup.

When this goes in I'll have a look at the "toggle breaks if scrollbar appears in-page" issue, and any Firefox-specific refinements you think remain (I don't use it myself).

ajayyy:
Thanks!

PR #1242, opened 2022-03-20
Improve "load in Firefox" docs

AlecRust:
There seems to be a difference in how you need to load `/dist` contents in to Firefox vs Chrome i.e. in Chrome you can load the "unpacked" folder itself, in Firefox you need to load a ZIP.

I bumped in to this when following the instructions, so this updates those instructions to clarify the difference.

***


ajayyy:
thanks

PR #1250, opened 2022-03-27
Improve the channel ID lookup.

### LINKED ISSUE #1249: channelID lookup optimizations
- [x] currently working on this

cc @zedseven

zedseven:
Closes #1249. @mchangrh 

This PR:
- Fixes desktop YouTube channel ID lookup. Previously it would grab entries from the Home tab which could sometimes be open in the background, and this also caused channel custom URLs to sometimes be grabbed as well.
- Potentially makes the Invidious lookup more stable.
- Makes the Regex more strict and reliable.

The reason the Invidious lookup uses `.channel-profile #channel-name` instead of just `#channel-name` is because the YouTube desktop client also uses `#channel-name`, but has over 50 instances of it. The `.channel-profile` makes sure the Invidious selector cannot accidentally match the wrong element in the YouTube desktop client. (if for example the chained selectors were converted to a promise race in the future)

***

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -1120,11 +1120,12 @@ function updatePreviewBar(): void {
 async function whitelistCheck() {
     const whitelistedChannels = Config.config.whitelistedChannels;
 
-    const getChannelID = () => videoInfo?.videoDetails?.channelId
-        ?? document.querySelector(".ytd-channel-name a")?.getAttribute("href")?.replace(/\/.+\//, "") // YouTube
-        ?? document.querySelector(".ytp-title-channel-logo")?.getAttribute("href")?.replace(/https:\/.+\//, "") // YouTube Embed
-        ?? document.querySelector("a > .channel-profile")?.parentElement?.getAttribute("href")?.replace(/\/.+\//, "") // Invidious
-        ?? document.querySelector("a.slim-owner-icon-and-title")?.getAttribute("href")?.replace(/\/.+\//, ""); // Mobile YouTube
+    const getChannelID = () =>
+        (document.querySelector("a.ytd-video-owner-renderer") // YouTube
+        ?? document.querySelector("a.ytp-title-channel-logo") // YouTube Embed
+        ?? document.querySelector(".channel-profile #channel-name")?.parentElement.parentElement // Invidious
+        ?? document.querySelector("a.slim-owner-icon-and-title")) // Mobile YouTube
+            ?.getAttribute("href")?.match(/\/channel\/(UC[a-zA-Z0-9_-]{22})/)[1];
 
     try {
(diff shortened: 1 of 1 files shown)


mchangrh (approved):
LGTM

test links
[embed](https://www.youtube.com/embed/videoseries?list=PLwstYNm1FaXfd1UxRiMjhZe8veSyIj67W)
[mobile](https://m.youtube.com/watch?v=_BnxWJjd8GI&persist_app=1&app=m)
[invidious](https://invidious.snopyta.org/watch?v=_BnxWJjd8GI)
[desktop](https://www.youtube.com/watch?v=_BnxWJjd8GI&persist_app=1&app=desktop)


* mobile link will force mobile layout, go to http://www.youtube.com/?persist_app=1&app=desktop to revert, desktop link vice versa

src/content.ts
```diff
@@ -1120,11 +1120,12 @@ function updatePreviewBar(): void {
...
 
-    const getChannelID = () => videoInfo?.videoDetails?.channelId
-        ?? document.querySelector(".ytd-channel-name a")?.getAttribute("href")?.replace(/\/.+\//, "") // YouTube
-        ?? document.querySelector(".ytp-title-channel-logo")?.getAttribute("href")?.replace(/https:\/.+\//, "") // YouTube Embed
-        ?? document.querySelector("a > .channel-profile")?.parentElement?.getAttribute("href")?.replace(/\/.+\//, "") // Invidious
-        ?? document.querySelector("a.slim-owner-icon-and-title")?.getAttribute("href")?.replace(/\/.+\//, ""); // Mobile YouTube
+    const getChannelID = () =>
+        (document.querySelector("a.ytd-video-owner-renderer") // YouTube
+        ?? document.querySelector("a.ytp-title-channel-logo") // YouTube Embed
+        ?? document.querySelector(".channel-profile #channel-name")?.parentElement.parentElement // Invidious
```
ajayyy, line 1126:
```suggestion
        ?? document.querySelector("a > .channel-profile > #channel-name")?.parentElement.parentElement // Invidious
```

zedseven: The `onInvidious` check looks good, but iirc the `#channel-name` element wasn't always an immediate child of `.channel-profile`, and that wasn't always an immediate child of an `a` tag.

If you can verify that is indeed the case, then it makes sense to include this change. I can't check it until tomorrow night at the earliest.

ajayyy: If that's not the case, I don't think we can rely on `.parentElement.parentelement`

zedseven: Ah, you're right. This should be merged then - I forgot about the `parentElement`.

Just wanted to make sure there was justification for the direct-child selector.

ajayyy:
thanks!

PR #1259, opened 2022-04-01
bump dependencies

mchangrh:

***

- update dependences
- don't update react b/c of breaking changes
- downgrade webpack to reduce one security risk
- change over to `fork-ts-checker-webpack-plugin` to reduce build times by 1/3


ajayyy:
Nice, so this runs a type checker in another process that prints errors, but still allows it to build successfully?

mchangrh:
pretty much, it basically only checks for errors once the fork has completed instead of inline

Issue #1275, opened 2022-04-09
Add warning to popup if beta server is enabled


MRuy:
Add a warning to the top of the popup to indicate that user has the beta server option enabled

MRuy:
.take

PR #1276, opened 2022-04-09
Add beta server warning to popup

MRuy:
#1275


ajayyy:
Reason for using toggle instead of remove? Shouldn't it always be removed?

MRuy:
I was thinking about toggling the class if options change and the popup is in the page active right now, but that needs observing the config for changes. No, removing is fine.

ajayyy:
image

PR #1293, opened 2022-04-23
Fix vote button hover inconsistency

AlecRust:
### Current Problem

If you hover over a segment's voting buttons when the popup is rendered in-page, they slightly reduce in opacity.

But if you hover over the voting buttons when the popup is rendered in the dropdown, they do not change in opacity.

This difference in behaviour based on where the popup is rendered looks like a bug to me.

### Proposed Solution

Fix this by removing `.voteButton` CSS from `content.css` and adding the missing `.voteButton:hover` CSS to `popup.css`

Also:
- Prefer `opacity` over `filter` for changing opacity

### Screenshots

Before             |  After
before  |  after

***

### DIFF
--- a/public/content.css
+++ b/public/content.css
@@ -9,7 +9,7 @@
     position: absolute;
     width: 100%;
 	pointer-events: none;
-	
+
 	height: 100%;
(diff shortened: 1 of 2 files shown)


ajayyy:
thanks!

PR #1330, opened 2022-05-24
Use Navigation API when available

### LINKED ISSUE #1328: Improve efficiency
SponsorBlock background currently registers `tabs.onUpdated` listener just to check for URL updates on two domains it is interested in (YouTube and YouTube no-cookie sites). This is an anti-pattern that is known to cause excessive background activations.
https://github.com/ajayyy/SponsorBlock/blob/024480513c7fc02b91eb9a90bad95de4926044ad/src/background.ts#L26

Fortunately, Chrome 102 (which is shipping today) has a much better alternative: [Navigation API](https://chromestatus.com/feature/6232287446302720). This API will actually allow extension to register listener in the content script directly (and avoid background context activation at all.) If you are interested, I can prepare a PR with this change. Also, (on YouTube specifically) older versions of Chrome and current versions of Firefox and Safari can have similar efficiency gains by registering listeners for YouTube-specific DOM events.

If you are interested, I can elaborate and prepare a PR.

bershanskiy:

***

Fixes #1328.

Tested this change on Chrome 102 and Firefox (I do not own any Apple device).

### DIFF
--- a/src/background.ts
+++ b/src/background.ts
@@ -23,14 +23,35 @@ if (utils.isFirefox()) {
     });
 } 
 
-chrome.tabs.onUpdated.addListener(function(tabId) {
-	chrome.tabs.sendMessage(tabId, {
+function onTabUpdatedListener(tabId: number) {
+    chrome.tabs.sendMessage(tabId, {
         message: 'update',
-	}, () => void chrome.runtime.lastError ); // Suppress error on Firefox
-});
+    }, () => void chrome.runtime.lastError ); // Suppress error on Firefox
+}
 
-chrome.runtime.onMessage.addListener(function (request, sender, callback) {
-	switch(request.message) {
+function onNavigationApiAvailableChange(changes: {[key: string]: chrome.storage.StorageChange}) {
+    if (changes.navigationApiAvailable) {
+        if (changes.navigationApiAvailable.newValue) {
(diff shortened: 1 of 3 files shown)


ajayyy:
Might be a good idea to future proof this by doing the check in the content script, setting a variable in local storage and using that to determine if tabs.onUpdate is needed

ajayyy:
I think 2 is fine, as it doesn't hurt to have two listeners running. So, the first load will have two listeners running, and subsequent page loads will only run one listener.

bershanskiy:
> There is a proxy to local storage already in config.js that can be used. `Config.config.historyApiAvailable`

I'm not sure I follow. Could you elaborate, please? Are you suggesting bundling this technical info together with real user-controlled data?

bershanskiy:
Also, I tested that indeed this patch indeed makes background context more efficient:
 - install both the current release extension and build with this patch
 - look at `chrome://extensions/` and wait until both extension's backgrounds go inactive
 - open a new tab (or navigate somewhere in an old tab or do anything else that would fire `tabs.onUpdated`
 - notice that modified extension's background is still inactive, while release extension's background goes active

bershanskiy:
@ajayyy Should I do anything to advance this PR?

PR #1331, opened 2022-05-25
update dependencies & bump minimum node version

### LINKED ISSUE #1329: Should minimum NodeJS version be bumped a bit?
`package.json` specifies that NodeJS 12.20.0 is supported:
https://github.com/ajayyy/SponsorBlock/blob/4738c1897f41c307b2a4643e214f2b95e1365add/package.json#L68

but codebase uses optional chaining operator `?.` which was first introduced in NodeJS 14.0.0 ([source](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#browser_compatibility)):
https://github.com/ajayyy/SponsorBlock/blob/4738c1897f41c307b2a4643e214f2b95e1365add/webpack/webpack.common.js#L93

There are some other minor nits in the codebase like dependencies that require NodeJS version no older than 12.22.0.

Also, CI seems to use NodeJS 16.

mchangrh:
- bump all dependences
- add github-actions reporter to jest
- default type converted to module for webpack
- convert (most) webpack configs to ESM
  - manifest.cjs cannot be converted since there is no native `.json` import until node 17

closes #1329

## React 18
React v18 is here and it comes with major changes that I'm not comfortable making
https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html

## npm audit
fixing the vulns means downgrading web-ext and webpack, which means introducing other vulns that are fixed by upgrading web-ext and webpack, you get the picture.


ajayyy:
thanks!

PR #1333, opened 2022-05-25
Debounce setting bar color

Argn0:
- [ ] I agree to license my contribution under LGPL-3.0 **or** my contribution is from another project with a license compatible with LGPL-3.0

***

I noticed that you could hit the chrome storage call limit very easily when setting the colors.


Argn0:
I changed it to 50 milliseconds. It works perfectly fine like this and makes it practically impossible to lose a change by closing the tab.

Do you think it's fine like this?

PR #1348, opened 2022-06-06
config diff plugin

mchangrh:
add config diff plugin to throw a warning if `config.json` differs from `config.json.example`

This is helpful if there's an change/update to the category list


ajayyy:
Maybe ignore server url and mention the specific part changed

mchangrh:
nvm

PR #1370, opened 2022-06-24
Show connectionError string if status is not 404

### LINKED ISSUE #1364: Show response status in popup for 500 instead of "no segments found"
Mentioned before in the discord but giving a unique error message on server errors instead of "no segments found" would be helpful so users don't run up into 5xx errors or rush to submit segments that are rejected after the fact

mchangrh:
show connectionError string if status is not 404

closes #1364

***


mchangrh:
the empty 200 results in false positives when none of the video hashes match, working on it

ajayyy:
What cases are blank 200s returned?

mchangrh:
> What cases are blank 200s returned?

it would return blank 200 on testing or if there are no segments on the video but the server returned responses for adjacent hashes

PR #1422, opened 2022-07-29
allow disabling mute segments locally

### LINKED ISSUE #1343: Allow "disabling" mute segments locally
Mute segments cannot  be hidden in the popup

mchangrh:

***

closes #1343


ajayyy:
I wonder what the original reasoning was

mchangrh:
My guess is that it was to exclude POI segments?

ajayyy:
seems weird to exclude those too

mchangrh:
let me test it again, I don't think it would be any issue with local disabling

PR #1423, opened 2022-07-29
don't show unskip first for full size notices

### LINKED ISSUE #1421: Manual skip won't work as it should when using mouse
When using `manual skip` the notice window that show up will `Unskip` instead of `Skip`.
image

So, if i want to skip i first have to press `Unskip` and then `Reskip`
image

It seems that if the `Unskip` is there (as in the upper picture) and i press `Enter` on the keyboard it works, but not if i use the mouse and click in the window.

SponsorBlock 4.7.1
Firefox 103.0
Chrome 103.0.5060.134

I also noticed that when i use the keyboard and press on the key ENTER the `Unskip (Enter)` in the window won't change between `Unskip (Enter)` and `Reskip (Enter)` as i do when i use the mouse. It will always show `Unskip (Enter)` even if it do switch between unskip and reskip in the video.

mchangrh:

***

closes #1421 

doesn't seem to be done intentionally, likely just an overlooked regression

### DIFF
--- a/src/components/SkipNoticeComponent.tsx
+++ b/src/components/SkipNoticeComponent.tsx
@@ -265,7 +265,7 @@ class SkipNoticeComponent extends React.Component<SkipNoticeProps, SkipNoticeSta
 
                 {/* Unskip/Skip Button */}
                 {!this.props.smaller || this.segments[0].actionType === ActionType.Mute
-                    ? this.getSkipButton(1) : null}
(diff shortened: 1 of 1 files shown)


ajayyy:
Don't really understand this solution. Now buttonIndex is unused and always 0. The reason they are handled in different places is to support 2 appearing at once when mute notices happen (unmute and unskip).

I think the bug is somewhere else

PR #1450, opened 2022-08-20
Tidy/simplify popup

AlecRust:
### Current Problem

Looks like lots of tweaks have happened to the popup, but there's various alignment issues, invalid HTML such as missing `alt` attribute, use of old `align` attribute etc.

### Proposed Solution

Somewhat tidy up the popup HTML and CSS. Improves alignment/padding, reduces unnecessary flexbox usage, ~fixes "Your Work" not having equal columns~ etc.

### Screenshots

Before             |  After
before  |  after

### DIFF
--- a/public/_locales/en/messages.json
+++ b/public/_locales/en/messages.json
@@ -122,6 +122,9 @@
     "closePopup": {
         "message": "Close Popup"
     },
+    "closeIcon": {
+        "message": "Close Icon"
+    },
     "SubmitTimes": {
         "message": "Submit Segments"
     },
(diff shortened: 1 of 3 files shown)


ajayyy:
The "Your Work" thing was on purpose since usernames can be long

AlecRust:
OK, I've reverted the 50/50 columns and improved the width when editing username.

Before/after screenshots have been updated.

One for a separate PR, but I feel like the username area is the most gnarly of this popup:
- Columns not being equal width looks like a bug. Especially for users with short usernames and many submissions.
- Left column is wider to accommodate more text, but also includes two buttons on that row
- Some fiddly/flaky CSS to accommodate all this including the inline edit form

PR #1454, opened 2022-08-21
Improve copy of start/end segment button

AlecRust:
### Current Problem

Ever since using this (excellent) extension the labelling of the main "Segment Starts Now" button has irked me 🙂 

Button labels should typically start with a verb i.e. the "action" that will be performed - "Start" in this case. Right now in both "states" of the button, the label starts with a noun - "Segment" in this case.

User shouldn't need to read somewhere in the middle of the button to determine the "state" the button is in.

### Proposed Solution

- `Segment Starts Now` -> `Start Segment Now`
- `Segment Ends Now` -> `End Segment Now`

### Screenshots

Before             |  After
before  |  after

### DIFF
--- a/public/_locales/en/messages.json
+++ b/public/_locales/en/messages.json
@@ -77,10 +77,10 @@
         "message": "No segments found"
     },
     "sponsorStart": {
-        "message": "Segment Starts Now"
+        "message": "Start Segment Now"
     },
     "sponsorEnd": {
-        "message": "Segment Ends Now"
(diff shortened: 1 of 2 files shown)


AlecRust:
One for a separate PR, but due to the importance of these start/stop buttons I feel like they could benefit from being full width, and include an icon for each state to further differentiate.

ajayyy:
Looks good!

Todo for me: trigger an update in crowdin instead of replacement before merging

PR #1521, opened 2022-10-08
Update popup on segment updates + some code cleanup

mini-bomba:

***
This PR:
* Triggers a segment list refresh in the popup when content script loads new ones
* Changes the return value of `getYouTubeVideoID` on fail to `null` (simplifies type annotations & logic)
* Deduplicates and cleans up popup -> content script communication code in `popup.ts`
  * Removes the `sendMessage` function
  * Modifies `sendTabMessage` to use a provided callback, adds `sendTabMessageAsync` that returns a promise
  * Modifies any function that duplicated `sendTabMessage` code to call `sendTabMessage` or `sendTabMessageAsync`, making them async if needed and adding type annotations for response objects

### DIFF
--- a/src/background.ts
+++ b/src/background.ts
@@ -2,13 +2,13 @@ import * as CompileConfig from "../config.json";
 
 import Config from "./config";
 import { Registration } from "./types";
-import Utils from "./utils";
-import { GenericUtils } from "./utils/genericUtils";
 
 // Make the config public for debugging purposes
 
 window.SB = Config;
 
+import Utils from "./utils";
+import { GenericUtils } from "./utils/genericUtils";
 const utils = new Utils({
     registerFirefoxContentScript,
     unregisterFirefoxContentScript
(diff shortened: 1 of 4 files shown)


src/background.ts
```diff
@@ -2,13 +2,13 @@ import * as CompileConfig from "../config.json";
...
 import { Registration } from "./types";
+import Utils from "./utils";
+import { GenericUtils } from "./utils/genericUtils";
 
 // Make the config public for debugging purposes
 
 window.SB = Config;
 
-import Utils from "./utils";
-import { GenericUtils } from "./utils/genericUtils";
```
ajayyy, line 11: Please undo this

src/content.ts
```diff
@@ -441,7 +441,7 @@ async function videoIDChange(id): Promise<void> {
     //close popup
     closeInfoMenu();
```
ajayyy, line 442: If we are calling update on the popup, can this be removed?

ajayyy (changes requested):

src/content.ts
mini-bomba: Did some testing, seems like this can indeed be safely removed now.
The popup correctly persists and updates between video changes (manual, playlist, autoplay and even going to the home/search page)

Issue #1537, opened 2022-10-14
More strict ts


ajayyy:
noFallthroughCasesInSwitch
noImplicitReturns

More work:
strictNullChecks

mchangrh:
.take

ljkGH:
can be closed since it's merged?

PR #1550, opened 2022-10-21
add noFallThrough, inplicitReturn, update packages

mchangrh:

***

#1537 
no strictNullChecks as it's a massive undertaking that requires some rewriting


src/background.ts
```diff
@@ -59,16 +59,16 @@ chrome.runtime.onMessage.addListener(function (request, sender, callback) {
...
-            return;
+            return true;
         case "openUpsell":
             chrome.tabs.create({url: chrome.runtime.getURL('upsell/index.html')});
-            return;
+            return true;
         case "openPage":
             chrome.tabs.create({url: chrome.runtime.getURL(request.url)});
-            return;
+            return true;
```
ajayyy, line 71: I think these should all be false, otherwise the port stays open

src/content.ts
```diff
@@ -2093,7 +2093,7 @@ async function voteAsync(type: number, UUID: SegmentUUID, category?: Category):
     const sponsorIndex = utils.getSponsorIndexFromUUID(sponsorTimes, UUID);
 
     // Don't vote for preview sponsors
-    if (sponsorIndex == -1 || sponsorTimes[sponsorIndex].source !== SponsorSourceType.Server) return;
+    if (sponsorIndex == -1 || sponsorTimes[sponsorIndex].source !== SponsorSourceType.Server) Promise.resolve();
```
ajayyy, line 2096: Will this still short-circuit the function?

src/popup.ts
```diff
@@ -67,9 +67,13 @@ class MessageHandler {
...
 let allowPopup = window === window.top;
-window.addEventListener("message", async (e) => {
+window.addEventListener("message", async (e): Promise<void> => {
     if (e.source !== window.parent) return;
-    if (e.origin.endsWith('.youtube.com')) return allowPopup = true;
+    if (e.origin.endsWith('.youtube.com')) {
+        allowPopup = true;
+        return
+    }
+    else return;
```
ajayyy, line 76:
```suggestion
    if (e.origin.endsWith('.youtube.com')) {
        allowPopup = true;
    }
```

I think this would still be allowed

src/background.ts
mchangrh: true/false is for async or sync

src/content.ts
mchangrh: well it's that or Promise.reject()

ajayyy: Shouldn't there still be a return though after resolving the promise. Otherwise it's going to run all the code underneath

src/background.ts
ajayyy: True = async

src/content.ts
mchangrh: oh oops

PR #1560, opened 2022-10-27
add selector for piped

### LINKED ISSUE #1243: Support submiting segments for piped frontend for youtube
I use [piped](https://github.com/TeamPiped/Piped) for watching youtube videos on the web (similar to invidious) and it would be really nice if i was able to submit segments while watching on piped (skipping segments already works).

mchangrh:

closes #1243

- add playerInit listener for piped
- add piped specific selectors
- explicitly add css files to web accessible resources

Working
  - Preview bar segments
  - videoID detection
  - skipping
  - skip notices
  - unskipping
  - marking start, end times with popup
  - submitting with popup

Not Working
  - in-page popup

### DIFF
--- a/manifest/manifest.json
+++ b/manifest/manifest.json
@@ -69,8 +69,11 @@
         "icons/PlayerInfoIconSponsorBlocker.svg",
         "icons/PlayerDeleteIconSponsorBlocker.svg",
         "popup.html",
+        "popup.css",
         "content.css",
-        "js/document.js"
+        "shared.css",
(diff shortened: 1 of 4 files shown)


mchangrh:
## THIS WILL NOT WORK ON ANY PIPED INSTANCES

This only currently works on a special piped instance with the playerInit event https://player-init-event.piped.pages.dev/

there is no checking so it will conflict with built in sponsorBlock settings, I will leave that up to the Piped Devs.

Issue #1570, opened 2022-11-02
Make a P2P network for SponsorBlock database


jcastro:
I LOVE this service and would like to contribute, I don't know how to developer but I do have some 24/7 servers that I could use to host an instance of the app? Would that be possible? Or somehow have the DB somewhere where we can all update it so the service would go barely offline

skeddles:
yes please. being able to host a local copy of the db would be nice too, or at least cache certain channels so it doesnt become useless when the server goes down.

ajayyy:
https://github.com/mchangrh/sb-mirror

jcastro:
@ajayyy this is awesome! thanks so much. I was wondering if it's possible to add several address to 'SponsorBlock Server Address' config in the extension settings? so we can have several mirrors added there as well

<img width="563" alt="image" src="https://user-images.githubusercontent.com/190036/199635410-4917aa06-d5a8-4123-a3b0-37485136ed2e.png">

PanoptesDreams:
> @ajayyy this is awesome! thanks so much. I was wondering if it's possible to add several address to 'SponsorBlock Server Address' config in the extension settings? so we can have several mirrors added there as well
> 
> <img alt="image" width="563" src="https://user-images.githubusercontent.com/190036/199635410-4917aa06-d5a8-4123-a3b0-37485136ed2e.png">

Create a separate issue as a feature request, instead of piggybacking on others.

lidel:
One way to do this, is to put the DB on [IPFS](https://ipfs.tech) (as immutable snapshots) and set up either [IPNS](https://docs.ipfs.tech/concepts/ipns/) and/or [DNSLink](https://dnslink.dev/) for publishing updates pointing at the latest version.

This not only allows P2P retrieval and makes it easier for people to co-host the DB, but also provides an HTTP CDN for regular browsers thanks to [public gateways](https://ipfs.github.io/public-gateway-checker/)

Example (DNSLink):
- If HTTP server hosting https://en.wikipedia-on-ipfs.org  is down, one can still load content P2P via IPFS, or use any public gateway as a fallback mirror: 
  - https://dweb.link/ipns/en.wikipedia-on-ipfs.org
  - https://cf-ipfs.com/ipns/en.wikipedia-on-ipfs.org (Cloudflare)
  
Going IPFS route has a nice property  of being backward and forward compatible:
- Regular browsers still benefit from having multiple mirrors that can be used as a fallback if the original server is down.
[...]

Issue #1562, opened 2022-10-27
Fallback server option


Timmyfox:
Considering how frequently the main server has been going down and having issues lately, perhaps the “SponsorBlock Server Address” setting could be be amended with fallback options?

This way, the user can specify one or more mirrors which can be used as fallback in the event that the primary server gives an error.

For example, let's say the following are specified:
- Default primary server `https://sponsor.ajay.app`
- Secondary server `https://example.foo.bar` (mirror)

Here it will first check the primary server just as before. But if there is an error returned in trying to retrieve the data, the extension can then check the secondary `example.foo.bar` server, which is acting as a mirror, and hopefully be able to retrieve the data from there instead.

One or two additional servers to be enough but more shouldn't hurt. The field(s) would probably be blank/disabled by default unless you wish to provide an official fallback mirror, my thinking is more to allow more savvy users to host or provide their own mirrors, or choose an existing 3rd party mirror, as a fallback instead of replacing the default server.

lewisdoesstuff:
I've just submitted PR #1572 that adds this, so you can build it from my branch in the meantime :)

ccuser44:
Even better.

If we had a few servers at different geographical locations and the extension would use the closest one.

This would be great for load balancing.

Like for example:

- Netherlands
- Poland
- USA
- Brazil
- Singapore
- India
- South Africa

PR #1572, opened 2022-11-03
Added fallback server option

lewisdoesstuff:
As the main server seems to be having some issues, I've added an option to use a fallback server when the main one fails. 

This defaults to "" (`undefined`) in the config, with users being able to set the URL below the standard URL option in options. 

If you've got `testingServer` set, this is overridden, and the form is removed from the options. 

How it works: 

Upon requesting sponsors, if the main server fails, we call `retryFetch` to try again. As we keep note of how many times we retry, we use the fallback server on every even retry, so it flips back and forth. 

The delay between retries has also been adjusted. 404 is the same, but if we've failed 2 or less times, we retry instantly, before going back to the previous delay setting.

As most users are only concerned about fetching sponsors, we don't use the fallback server for requests regarding submitting data.

We also only use the fallback server on `asyncRequestToServer()` calls, as nothing using the synchronous method would benefit.

As mentioned in issues #1570 and #1562 

***

### DIFF
--- a/config.json.example
+++ b/config.json.example
@@ -1,7 +1,9 @@
 {
     "serverAddress": "https://sponsor.ajay.app",
     "testingServerAddress": "https://sponsor.ajay.app/test",
+    "fallbackServerAddress": "",
     "serverAddressComment": "This specifies the default SponsorBlock server to connect to",
+    "fallbackAddressComment": "This specifies the server SponsorBlock will attempt to connect to if the primary one fails.",
     "categoryList": ["sponsor", "selfpromo", "exclusive_access", "interaction", "poi_highlight", "intro", "outro", "preview", "filler", "chapter", "music_offtopic"],
     "categorySupport": {
         "sponsor": ["skip", "mute", "full"],
--- a/public/_locales/en/messages.json
+++ b/public/_locales/en/messages.json
@@ -505,6 +505,12 @@
     "customServerAddressDescription": {
         "message": "The address SponsorBlock uses to make calls to the server.\nUnless you have your own server instance, this should not be changed."
     },
+    "fallbackServerAddress": {
+        "message": "SponsorBlock Fallback Server Address"
+    },
+    "fallbackServerAddressDescription": {
+        "message": "The address SponsorBlock uses to make calls to the server if the main server is down."
+    },
     "save": {
         "message": "Save"
     },
--- a/public/options/options.html
+++ b/public/options/options.html
@@ -554,6 +554,26 @@
 						</div>
 					</div>
 				</div>
+				
+				<div data-type="text-change" data-sync="fallbackServerAddress" data-dependent-on="testingServer" data-dependent-on-inverted="true">
+					<label class="optionLabel inline">
+						<span class="optionLabel">__MSG_fallbackServerAddress__:</span>
+	
+						<input class="option-text-box" type="text" style="margin-right:10px">
+					</label>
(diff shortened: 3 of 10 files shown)


mini-bomba:
There are some caveats with using mirrors that need to be kept in mind and accounted for.
- Currently, most mirrors only implement the bare minimum - just the GET /skipSegments/<hash> endpoint.
- Mirrors will not accept any submissions.
  - Also any submission action will leak the privateID to a mirror/fallback server.

From what I can see in the diff, your changes probably sometimes forward more than just GET /api/skipSegments (most notably the POST /api/skipSegments endpoint, which is the submission endpoint mirrors currently will not accept)

Also a question to think about: Should the fact that a fallback server is being used be indicated somewhere?

ajayyy:
I personally think it should be skip segments only

lewisdoesstuff:
I definitely agree, the fallback should only be used as a mirror to pull from, other requests should still use the main server. 

Let me make some changes to only have it on GET /api/skipSegments.

@mini-bomba - any ideas on what would work best for this? I wouldn't want something as intrusive as a browser alert every time it fetches from the fb. Potentially an alert when choosing to upload segments if we previously used the fallback server, or some static text in the dialog aside the video player? (I'm not sure what you'd call it).

lewisdoesstuff:
Well that broke my commit :/ 

Set fallbackServer to false on the POST /api/skipSegments endpoint
I'll have a think about implementing some kind of notification/status of which server's being used, as well as squash that commit history :)

ajayyy:
Probably makes most sense to have that in the popup. Instead of "Segments found on this video" say "Segments found on this video from the backup server"

lewisdoesstuff:
May have accidentally reverted to the commit before and force pushed :/ 

Readded these, squashed commits. 

> Instead of "Segments found on this video" say "Segments found on this video from the backup server" 

This makes the most sense to me - Need some sleep but I'll get that done tomorrow hopefully.

goodness-from-me:
Maybe set some default fallback server? I don't think that average SponsorBlock user would search for fallback server by himself.

mini-bomba:
> Maybe set some default fallback server? I don't think that average SponsorBlock user would search for fallback server by himself.

Could perhaps have a list of mirrors sent by the server, have the extension remember it for a few days and pick a random server from the list every time it updates the list, as long as the user didn't modify it.
Though maybe a select list would be better for this, with "Pick randomly", "Don't use a fallback server" and "Custom server" options....

But anyways, we probably should first implement the basic feature, then add further enhancements.

ajayyy:
No default fallback server for now

lewisdoesstuff:
Just finishing up implementing this, but noticed some strange behaviour - not sure if anyone more familiar may be able to shed some insight? 

The fallback switch works fine when loading a video without the main server reachable, but after switching videos a few times (in the recommended pane), it seems to stop working.  
*upon further testing, it seems to fail on a new video as well, sometimes*
The request to get the segments seems to be passed to `utils.asyncRequestToServer` and then to `utils.asyncRequestToCustomServer` with no issue, but it never gets a response from the listener in `background.ts`, nor the `sendRequestToCustomServer` in there. 

It looks like it'll occasionally throw 
```js
Unchecked runtime.lastError: A listener indicated an asynchronous response by returning true, but the message channel closed before a response was received
``` 
at some point, but this happens on a successful retry request as well.

PR #1577, opened 2022-11-06
Don't overwrite endcards time upon submitting

caneleex:

***


ajayyy:
hmm, does this occur if edit was never pressed?

caneleex:
> hmm, does this occur if edit was never pressed?

if you mean literally pressing the button, yes, you can also just click on the end time, change the timing, refresh, submit and it will occur

PR #1611, opened 2022-12-03
Adjust "Your Work" element and add new metric

ankit-patel-4710:

***
For my CMSC389T class, I have tried to make some improvements to this already great codebase and plugin. I tried my best to understand what is going on currently and add to it. I made 3 changes, two of which were stylistic and the other was adding another section to the "Your Work" element in the popup. The new section combines the total number of segments you've skipped with the number that you've helped other skip.

### DIFF
--- a/public/_locales/en/messages.json
+++ b/public/_locales/en/messages.json
@@ -153,7 +153,7 @@
          "message": "Submissions"
     },
     "savedPeopleFrom": {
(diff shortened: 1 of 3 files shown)


mchangrh:
What is the point of the metric for skipped + saved? They can be wildly varying and would likely also require an additional metric for time saved others + time skipped for self

PR #1623, opened 2022-12-17
Show Full-Video Labels on thumbnails

mini-bomba:

***
The /videoLabels endpoint was added to SponsorBlockServer quite a while ago, but I didn't see any progress on the extension side, so I decided to try implementing it myself.
Labels are shown as a circle of the category's color with the sponsorblock logo in the middle (screenshot below)
image
When you hover over the icon, it is expanded to a rounded rectangle, with the category name shown.
image
The text color can be white or black, depending on the category color. (this logic was copied from regular fv labels) The logo's color will also update. (yet another screenshot below)
image
Category colors will update live as you change your settings. This is done using CSS variables. (we could probably replace color logic in regular fv labels and the timeline to use those)

Label fetching & API data processing is done on the background page instead of the content script. This allows me to keep a cache of requested hash-prefix chunks to limit the amount of requests. Responses are cached for 1 hour. (or until browser shutdown) 
Requests use hash prefixes. 4 characters seems to be a bit low for labels - at most 2 videos seem to be returned. If hash prefixes are shortened, the amount of requests will be reduced, as this patch caches whole hash-prefix blocks instead of individual videos. I'd even consider having all full-video segments returned in a single request an option - There are only 47681 labels in the DB, and this only needs videoIDs and categories to function.

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -45,6 +45,7 @@ import { getHash, HashedValue } from "@ajayyy/maze-utils/lib/hash";
 import { generateUserID } from "@ajayyy/maze-utils/lib/setup";
 import { setThumbnailListener, updateAll } from "@ajayyy/maze-utils/lib/thumbnailManagement";
 import { labelThumbnails, setupThumbnailPageLoadListener } from "./utils/thumbnails";
+import * as documentScript from "../dist/js/document.js";
 
 const utils = new Utils();
 
@@ -111,7 +112,8 @@ setupVideoModule({
         updatePreviewBar();
         updateVisibilityOfPlayerControlsButton();
     },
-    resetValues
+    resetValues,
+    documentScript
 }, () => Config);
 setThumbnailListener(labelThumbnails);
 setupThumbnailPageLoadListener();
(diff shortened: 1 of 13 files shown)


ajayyy:
I wonder if it'll look nice slightly translucent

src/background.ts
```diff
@@ -253,3 +257,49 @@ async function sendRequestToCustomServer(type: string, url: string, data = {}) {
 
     return response;
 }
+
+async function getVideoLabel(videoID: VideoID): Promise<Category> {
+    const hashPrefix = (await utils.getHash(videoID, 1)).slice(0, 3);
```
ajayyy, line 262: If 3 is kept, server needs updates before merging (for cache)

src/content.ts
```diff
@@ -2484,6 +2460,82 @@ function addPageListeners(): void {
...
+}
+
+function setUpThumbnailObserver() {
+    if (thumbnailObserver) {
+        thumbnailObserver.disconnect();
+        thumbnailObserver = null;
+    }
+    if (Config.config.fullVideoSegments) {
+        thumbnailObserver = new MutationObserver(thumbnailObserverHandler);
+        thumbnailObserver.observe(document, {childList: true, subtree: true})
```
ajayyy, line 2473: This is going to be called for _every_ child list change. I wonder if there's a better option here

mini-bomba: Probably the best option would be to somehow hook into the thumbnail custom element...

ajayyy:
General changes:

- Move cache out of background script. Background script is non persistent on chrome, and will be on all browsers on Manifest v3. This makes it live less long than the browser tab

- Switch document script injection system to work on channel pages by injecting raw html instead of using a URL

- Use mutation listener on thumbails to detect changing related videos and switching from recently uploaded to popular on channels

- Changed cache invalidation

***

I still have a bit to finish before merging, but it looks awesome thanks so much!

PR #1627, opened 2022-12-27
highlight partial required segments

mchangrh:

***

as detailed in https://github.com/ajayyy/SponsorBlockServer/pull/412, partial hashes are possible but aren't correctly highlighted on the seekbar since the check is strict

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -1421,6 +1421,7 @@ function updatePreviewBar(): void {
     if (video === null) return;
 
     const hashParams = getHashParams();
+    const requiredSegment = hashParams?.requiredSegment as SegmentUUID || undefined;
     const previewBarSegments: PreviewBarSegment[] = [];
     if (sponsorTimes) {
         sponsorTimes.forEach((segment) => {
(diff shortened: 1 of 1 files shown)


src/content.ts
```diff
@@ -1434,7 +1435,7 @@ function updatePreviewBar(): void {
                 showLarger: segment.actionType === ActionType.Poi,
                 description: segment.description,
                 source: segment.source,
-                requiredSegment: hashParams.requiredSegment && segment.UUID === hashParams.requiredSegment
+                requiredSegment: requiredSegment && segment.UUID === requiredSegment || segment.UUID.startsWith(requiredSegment)
```
ajayyy, line 1438:
```suggestion
                requiredSegment: requiredSegment && segment.UUID.startsWith(requiredSegment)
```

mchangrh: would the `===` comparison not be faster? or at that point would it be better to just omit first and startWith()

ajayyy: yea I guess. Then can you add brackets to make operator presence clear?

Issue #165, opened 2019-11-05
Manifest v3 preparations


ajayyy:
https://developer.chrome.com/extensions/migrating_to_manifest_v3

> Are you making CORS requests in content scripts?
> 
> Move these requests to the background service worker.

https://developer.chrome.com/extensions/migrating_to_service_workers (might already be done)

ajayyy:
Fixed in 2.0

ajayyy:
Looks to be more work, but we will have to wait until it's implemented in Firefox

thipokKub:
Hello, chrome will deprecated Manifest v2 by January 2023 (and Firefox, sometime in 2023). So I am wondering is there any plan for the upcoming few months?

ajayyy:
I'm not too worried as this year's April fool's joke fork of this extension had to be in manifest v3, and I was able to do that in an hour, but it did involve disabling some things not needed for the joke.

trcrsired:
Will sponsorblock continue working in V3?

PR #1688, opened 2023-03-02
Change for the v3 manifest

judemont:

Hello,

I have modified the current version of the SponsorBlock extension by updating the manifest to support version 3. I made this change because Google has announced that version 2 of the Chrome extension API will no longer be supported starting in March 2023.

I forked the original project and made the necessary changes to the manifest. I also verified that the extension works properly with this new version of the manifest.

I am proposing this pull request so that you can review the changes and integrate them into the original project. If you have any comments or suggestions, please let me know.

Thank you for your attention.

### DIFF
--- a/manifest/chrome-manifest-extra.json
+++ b/manifest/chrome-manifest-extra.json
@@ -1,8 +1,5 @@
 {
   "optional_permissions": [
     "declarativeContent"
-  ],
-  "background": {
-    "persistent": false
-  }
+  ]
 }
--- a/manifest/firefox-beta-manifest-extra.json
+++ b/manifest/firefox-beta-manifest-extra.json
@@ -1,8 +1,8 @@
 {
-    "browser_specific_settings": {
-        "gecko": {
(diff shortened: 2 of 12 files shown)


ajayyy:
FYI there is no date for manifest v3, it is postponed indefinitely https://developer.chrome.com/docs/extensions/mv3/mv2-sunset/

judemont:
I apologize for my initial confusion regarding the support for manifest v3. Thank you for letting me know that the migration is postponed indefinitely.

However, I would like to emphasize that even though the migration date is still unknown, it is important to prepare in advance for this transition. Extensions that do not account for the changes introduced by manifest v3 may stop functioning correctly or become unavailable to users in the future.

ajayyy:
But upgrading too early breaks the extension for the many users using older browser versions

judemont:
Okay, no problems.

mchangrh:
just as a reference #165 start talking about mv3

PR #1710, opened 2023-03-24
Don't attach controls to inline preview players

mini-bomba:

***
This should fix an issue where starting your youtube session from the home page would cause SB controls to attach to the wrong player. (unless you used :selfpromo: yt-neuter)

### DIFF
--- a/src/utils/pageUtils.ts
+++ b/src/utils/pageUtils.ts
@@ -14,14 +14,18 @@ export function getControls(): HTMLElement {
     ];
 
     for (const controlsSelector of controlsSelectors) {
-        const controls = document.querySelectorAll(controlsSelector);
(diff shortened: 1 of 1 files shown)


src/utils/pageUtils.ts
```diff
@@ -14,16 +14,24 @@ export function getControls(): HTMLElement {
...
 
     return null;
 }
 
+export function isInPreviewPlayer(element: Element): boolean {
+    while (element != null) {
+        if (element.id === "inline-preview-player") return true;
+        element = element.parentElement;
+    }
+    return false;
```
ajayyy, line 32:
```suggestion
    return !!element.closest("#inline-preview-player");
```

let me know if that works

mini-bomba: It obviously does.
i love not knowing the whole std library, then reimplementing simple things like these and finally realizing i've written redundant code

PR #1711, opened 2023-03-25
Reset fvlabel color definitions when segment == null

mini-bomba:

***
Without this, we would try to use the color for an 'undefined' category on the first render pass of the element. It was then immediately re-rendered with a segment set, but DR missed the update, which caused it to stick to that 'undefined' category and the label became transparent.

In short, this is yet another fix/workaround for a DR bug

### DIFF
--- a/src/components/CategoryPillComponent.tsx
+++ b/src/components/CategoryPillComponent.tsx
@@ -43,7 +43,7 @@ class CategoryPillComponent extends React.Component<CategoryPillProps, CategoryP
 
         return (
             <span style={style}
-                className={"sponsorBlockCategoryPill"} 
+                className={"sponsorBlockCategoryPill"}
                 aria-label={this.getTitleText()}
                 onClick={(e) => this.toggleOpen(e)}
                 onMouseEnter={() => this.openTooltip()}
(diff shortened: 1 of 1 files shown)


src/components/CategoryPillComponent.tsx
```diff
@@ -117,13 +117,13 @@ class CategoryPillComponent extends React.Component<CategoryPillProps, CategoryP
     private getColor(): string {
         // Handled by setCategoryColorCSSVariables() of content.ts
         const category = this.state.segment?.category;
-        return `var(--sb-category-preview-${category}, var(--sb-category-${category}))`;
+        return category == null ? null : `var(--sb-category-preview-${category}, var(--sb-category-${category}))`;
```
beingfranklin, line 120: I would personally use nullish coalescing operator (??) in the code:
       ` return var(--sb-category-text-preview-${category ?? ''}, var(--sb-category-text-${category ?? ''})); `       
Suggesting this approach as its cleaner and more concise. However, if you have a specific reason for using the current implementation, please feel free to share it. :)

mini-bomba: Sorry, but I don't think this helps fix the bug. The issue wasn't some kind of nullpointer error (ts equivalent of that course). The code already generated rules such as `--sb-category-preview-undefined`. Your code would essentially do the same thing, just generating `--sb-category-preview-` instead. The issue was that such CSS variables aren't defined by the CSS variable generator in content.ts and that this property was quickly changed to a valid/defined one on first initialization of the category pill, but dark reader wasn't able to catch that later update.

Making sure the whole property is null when category is null (therefore removing it from inline css) ensures there is only a single color/backgroundColor property update, which leaves less room for dark reader to screw up.

beingfranklin: Okay, makes sense 👍

PR #1715, opened 2023-04-11
Add notice time to translation file

Blueberryy:
- [ x] I agree to license my contribution under LGPL-3.0 **or** my contribution is from another project with a license compatible with LGPL-3.0

***

### DIFF
--- a/src/components/NoticeComponent.tsx
+++ b/src/components/NoticeComponent.tsx
@@ -195,7 +195,7 @@ class NoticeComponent extends React.Component<NoticeProps, NoticeState> {
                         id={"skipNoticeTimerText" + this.idSuffix}
                         key="skipNoticeTimerText"
                         className={this.state.countdownMode !== CountdownMode.Timer ? "hidden" : ""} >
-                            {chrome.i18n.getMessage(NoticeTimeAfterSkip).replace("{seconds}", this.state.countdownTime)}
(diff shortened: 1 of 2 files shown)


Blueberryy:
don't know how to make this work, if anyone knows just give me snippet there

src/components/NoticeComponent.tsx
```diff
@@ -195,7 +195,7 @@ class NoticeComponent extends React.Component<NoticeProps, NoticeState> {
                         id={"skipNoticeTimerText" + this.idSuffix}
                         key="skipNoticeTimerText"
                         className={this.state.countdownMode !== CountdownMode.Timer ? "hidden" : ""} >
-                            {this.state.countdownTime + "s"}
+                            {chrome.i18n.getMessage("NoticeTimeAfterSkip").replace("{seconds}", this.state.countdownTime)}
```
mini-bomba, line 198: We're using TS, so the `replaceValue` (second argument) to `.replace()` needs to be a string, but `countdownTime` is a number. Adding `.toString()` should make TS happy. Alternatively you can wrap the whole number value in `String()` - not sure which one is better here.
```suggestion
                            {chrome.i18n.getMessage("NoticeTimeAfterSkip").replace("{seconds}", this.state.countdownTime.toString())}
```

Blueberryy: Thank you! I pushed your changes just now

ajayyy:
nice thanks

PR #1739, opened 2023-05-02
Show time with skips removed on invidious

### LINKED ISSUE #1717: Display "Time with skips removed" option on Invidious pages.
I understand that different invidious instances are different, and that showing the "time with skips removed" option may be difficult or impossible, so an alternative way to display this information could be showing it in the extension popup.

mini-bomba:

***
Almost no code changes were required to do this. Only difference between youtube and invidious is that we don't want to add any classes to the attached span element.

closes #1717

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -2375,14 +2375,15 @@ function updateAdFlag(): void {
 }
 
 function showTimeWithoutSkips(skippedDuration: number): void {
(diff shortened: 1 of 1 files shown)


ajayyy:
Screenshot?

mini-bomba:
Here's a screenshot of the controls bar after the change
screenshot

PR #1749, opened 2023-05-07
Set min width of previewbar segments to 1px

mini-bomba:

***
This fixes short segments on long videos not appearing at all in small embeds
[Issue reported on discord](https://discord.com/channels/603643120093233162/603643256714297374/1104742118519488643)

### DIFF
--- a/public/content.css
+++ b/public/content.css
@@ -45,6 +45,7 @@ div:hover > #previewbar.sbNotInvidious {
 .previewbar {
 	display: inline-block;
 	height: 100%;
+	min-width: 1px;
 }
 
 .previewbar.requiredSegment {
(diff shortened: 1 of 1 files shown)


ajayyy:
Does this cause the end time to be off, or the start time. Ideally we only want the end time to be offset probably.

mini-bomba:
Seems like this should cause end time to be shifted if required.
To confirm this, I performed a smol experiment:
a div, absolutely positioned near the center of a page: 20% of view height from top and bottom, 20% view width from left and 60% view width from right. This should result in a width of 20% view width, screenshot below
image
I then added a max-width property forcing the width to at least 40% view width. This resulted in the right edge being shifted.
image

ajayyy:
niiice perfect thanks

PR #1815, opened 2023-08-01
[docs] move building from readme, formatting tweaks

mchangrh:
- reformatted windows build issues
- updated contributing to reflect LGPL-3.0-or-later
- replaced Invidio.us with Invidious
- replaced API docs link
- updated Invidious API link


ajayyy:
I think the Firefox reviewers would prefer it in the readme

ajayyy:
Oh, there's a link, I guess that works

PR #1845, opened 2023-08-28
Fix duration time to consider playback rate

vicente-higino:

***

The duration of the video with skips now uses the playback rate to adjust the time.

Ex: 0:13 / 10:50 (9:16) @ 1x speed => 0:13 / 10:50 (4:38) @ 2x speed

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -843,8 +843,8 @@ function setupVideoListeners() {
         const rateChangeListener = () => {
             updateVirtualTime();
             clearWaitingTime();
-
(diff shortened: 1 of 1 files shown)


ajayyy:
I don't think this makes sense as a default. It could be an option though

mchangrh:
I don't think this makes sense at all unless you correct youtube's current and end times to also align with the time without segments - in which case it would be better suited for a plugin/userscript to parse instead for uniformity across all durations.

PR #1885, opened 2023-10-19
Refactored inspectTime into previewTime

Choromanski:

***

Looking through the source code I noticed that these two functions are extremely similar and thought there was two options to clean it up:
1. Merge the two functions into one [Implemented this one]
2. Simplify the old `previewTime` function by removing the `skipToEndTime` argument and implementing functionality for it in the `inspectTime` function


Choromanski:
Yeah it kills the readability, if you are looking for more readable code would something like [this](https://github.com/Choromanski/SponsorBlock/commit/f9200bd7847df90a6343b36cc660df4d977444fc) be preferred as Inspect and End buttons are functionally identical.

Issue #1821, opened 2023-08-10
Username Edit/Copy Overlapping 'Submissions' Section


Ci303:
**Description:**
When the browser window is resized to a width of 1078px, an issue occurs where the username edit/copy feature overlaps with the 'Submissions' section. This problem is observed while using a vertical 1080p monitor for video/chat purposes, with YouTube occupying the bottom half of the browser app on the screen.

**Steps to Reproduce:**
![username edit copy overlap](https://github.com/ajayyy/SponsorBlock/assets/36490238/9f53d2b5-f8a7-4916-afe0-a8d3a251d7ac)
![username edit copy overlap monitor](https://github.com/ajayyy/SponsorBlock/assets/36490238/f9fd54c1-9c1d-42c5-a4a4-d32095b9313a)

1. Open a web browser on a vertical 1080p monitor.
2. Navigate to a page with the username edit/copy feature and the 'Submissions' section.
3. Resize the browser window's width to 1078px.
4. Observe how the username edit/copy feature overlaps with the 'Submissions' section.

**Expected Behavior:**
The username edit/copy feature should not overlap with the 'Submissions' section even when the browser window is resized to a width of 1078px. Elements should be displayed properly, ensuring a user-friendly experience regardless of the window size.

**Actual Behavior:**
Upon resizing the browser window to 1078px in width, the username edit/copy feature overlaps with the 'Submissions' section, causing a visual and potentially functional conflict for users.

**Additional Information:**
- This issue primarily affects users who are utilizing a vertical 1080p monitor for video/chat activities.
- The problem specifically arises when YouTube is open in the bottom half of the browser app on the screen.
- Providing a responsive design that adjusts elements to different screen sizes could potentially address this problem and enhance user experience.

MaxS890:
**Suggested fix:**
Vertically stack the icons.

Issue #1429, opened 2022-08-04
"min-width: 50%" may cause #usernameElement to shrink


Black-Platypus:
With the Russian translation, for example, the `#sponsorTimesContributionsContainer>p.u-mZ.grey-text` has a long content, and  `#sponsorTimesContributionsContainer` is squishing `#usernameElement` together, even though there is a lot of empty space wasted.
This leads to undesired overflow that may leave the button(s) unclickable.
![image](https://user-images.githubusercontent.com/13470026/182759452-92ce3d6a-baa3-40e1-8e7f-6599dec2637f.png)

Weirdly, the `min-width` on `#usernameElement` seems to make it shrink as soon as the `#sponsorTimesContributionsContainer` becomes too wide, whereas not setting it will have it fall back to `auto`, which results in a better distribution of space among the elements, leaving `#usernameElement` **bigger**.
![Capture 2022-08-04 05-58-40 optimized](https://user-images.githubusercontent.com/13470026/182760561-71794a99-504c-4a9d-b2b0-ffe36c5d02c3.gif)
![Capture 2022-08-04 04-59-21](https://user-images.githubusercontent.com/13470026/182760580-10c56f29-869a-48f0-b95e-7d5664a019b3.png)

This would solve the above issue.
Who would have guessed that setting a _minimum width_ would result in something _shrinking_? 😅 Not I!

However, I don't know if this is the best solution and I can't test it in different environments right now, so I'm writing this up as an issue rather than creating a PR.
Maybe setting a `max-width` of 90px for the  `#sponsorTimesContributionsContainer>p.u-mZ.grey-text` element would be preferred, seeing as we're already using a pixel-based `max-width` on the left side?

PR #1895, opened 2023-10-24
Updated max-width to fix overlap error

Choromanski:

***

Solution for #1821 

Before: 
image

After:
image

### DIFF
--- a/public/popup.css
+++ b/public/popup.css
@@ -510,7 +510,7 @@
   white-space: nowrap;
   overflow: hidden;
   margin: 0 8px 0 0;
-  max-width: 165px;
+  max-width: 145px;
 }
 /*
  * Set username form container with "expanded" state


ajayyy:
How does this look with the popup in the browser top right

Choromanski:
Before: 

image


After: 

image

ajayyy:
Related #1429

Choromanski:
Issue #1429 results in the text running off the side of the page and can probably be closed:
image

Choromanski:
The discrepancy between youtube popup and browser popup is caused by the scrollbar, how do you feel about potentially removing the scrollbar but keeping scrolling functionality?

@ajayyy

ajayyy:
Hiding scrollbars is generally bad for accessibility, so I don't think it should be done. That may be why I never noticed it because I use Linux and have a very slim scrollbar

PR #1906, opened 2023-11-09
Create bug report template

rugk:
What about a basic bug report template?

Adjusted from https://github.com/rugk/offline-qr-code/blob/master/.github/ISSUE_TEMPLATE/bug_report.md

***

### DIFF
--- a/.github/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,41 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug
+assignees: ''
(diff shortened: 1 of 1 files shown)


mchangrh:
We might need some sanitization for debug since it contains unsubmitted videos and a big problem is browser variants (ff-variants specifically) so that would be helpful to include

otherwise LGTM except it's one `.github` folder too deep

rugk:
Okay so, did some changes and also clarified the browser variant thingy. @mchangrh anything else?

mchangrh:
LGTM

rugk:
Okay, I can only agree to disagree, because I personally find it useful to get better and more structured bug reports, but it's of course your thing/decision…

PR #1907, opened 2023-11-09
Add WebExt Lint as CI job

rugk:
Inspired by https://github.com/rugk/awesome-emoji-picker/blob/main/.github/workflows/webext-lint.yml

As far as I see you don't run that anywhere else, so may this be a good idea?

Same as https://github.com/ajayyy/DeArrow/pull/187

***

### DIFF
--- a/.github/workflows/.github/workflows/webext-lint.yml
+++ b/.github/workflows/.github/workflows/webext-lint.yml
@@ -1,0 +1,23 @@
+name: Web-Ext lint
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+  workflow_dispatch:
+
+jobs:
+  lint:
+    name: "Lint"
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: "web-ext lint"
+        uses: kewisch/action-web-ext@v1
+        with:
+          cmd: lint
+          source: src
+          channel: listed


ajayyy:
Warnings will need to be dealt with before merging.

mchangrh:
the v1 tag also seems to be built in 2019 which iirc is at least one breaking GHA version behind - should probably convert to locked commit

rugk:
Well dependabot can handle GitHub Actions and upgrade them (though I don't know if it automatically does as by your config)… then you can easily use version numbers instead of commits.

mchangrh:
dependabot hasn't triggered for our actions before and I don't trust it..., either way the tag is outdated and we've already set a precedence of locking by commit

rugk:
> dependabot hasn't triggered for our actions before and I don't trust it...

Well that is because you have not enabled it (respectively only for security updates), see https://github.com/ajayyy/DeArrow/pull/179. With a YAML configuration you can _exactly_ control what is updated and e.g. also only enable it for GitHub Actions. See https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates#overriding-the-default-behavior-with-a-configuration-file for an example.
If enabled, it is quite reliable in my experience, I mean it may take some time, but at least you notice updates.

.github/workflows/.github/workflows/webext-lint.yml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  lint:
+    name: "Lint"
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: "web-ext lint"
+        uses: kewisch/action-web-ext@v1
```
rugk, line 19:
```suggestion
        uses: kewisch/action-web-ext@cb8a694
```

Like this or what do you want here?

PR #2036, opened 2024-06-12
Fix refreshing

mini-bomba:

***
This PR fixes an issue that was reported on discord, where the after pressing the refresh button segments would sometimes show up and sometimes the segment list would be empty.
[Latest report/investigation thread in #concerns on discord](https://discord.com/channels/603643120093233162/603643256714297374/1250559680351174838)

This appears to be caused by a hidden iframe which holds the page `https://accounts.youtube.com/RotateCookiesPage`, where the content script would be injected in addition to the root page.
This iframe is only present when logged in, and as such this issue does not appear when logged out.
Both the root and iframe content scripts would subscribe to messages from the popup page, which caused a sort of race condition when the refresh button was pressed.
The iframe page's videoID would always be `null`. Due to a possible oversight, the `sponsorsLookup` function did not check whether videoID was valid, and ended up always trying to fetch the `7423` hashblock, which corresponds to the string `null`.
Depending on which content script got the data first, the segments list in the popup would appear or disappear.

This PR mitigates this issue by:
* Excluding the `https://accounts.youtube.com/RotateCookiesPage*` URL from content script's injection targets by adding an entry into the manifest
* Adding null checks before the `sponsorsLookup` function is called
* Adding a null check into the `sponsorsLookup` function which prints an error to the console if called with a null videoID and returns early

### DIFF
--- a/manifest/manifest.json
+++ b/manifest/manifest.json
@@ -11,6 +11,9 @@
             "https://*.youtube.com/*",
             "https://www.youtube-nocookie.com/embed/*"
         ],
+        "exclude_matches": [
+            "https://accounts.youtube.com/RotateCookiesPage*"
+        ],
         "all_frames": true,
         "js": [
             "./js/content.js"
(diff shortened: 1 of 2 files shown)


ajayyy:
nice

PR #2135, opened 2024-10-18
Fix skip audio notification not working on firefox

calledude:

***

Firefox seems to have issues playing an `.ogg` file, as it is recognizing it as a video (`video/ogg` MIME) file even though it has nothing but an audio stream.

image

image

It seems to be atleast _somewhat_ of a known bug. This PR merely renames the file and updates the manifest(s) to reflect that, which enables firefox to correctly identify which type of MIME type the file is and subsequently enables the sound to work properly again.
https://bugzilla.mozilla.org/show_bug.cgi?id=1240259


ajayyy:
Some people are reporting this change breaking audio notifications

PR #2162, opened 2024-11-29
Respect iOS/macOS accent color by using a monochrome toolbar icon for…

### LINKED ISSUE #1324: Consider respecting the system accent colour on Safari's toolbar icon
I'm not familiar with extension development for Safari, but it would seem it's possible to provide a non-colour icon and allow the system accent colour to show that the extension is active.

- [Response from Apple since this seems to be somewhat undocumented?](https://developer.apple.com/forums/thread/651500)
- Examples of the uniform look when extensions allow the accent colour to theme them:

strafe:
… Safari

***

This change would resolve #1324. To recap, Safari on macOS has a feature where active extensions are given the system's chosen accent colour, and inactive extensions are given a translucent appearance. Many (but not all) extensions support this, creating a consistent aesthetic.

For some reason this actually started working on earlier versions of macOS, my screenshot on the issue (https://github.com/ajayyy/SponsorBlock/issues/1324#issuecomment-1624176220) shows this, but as of Sequoia, the old behaviour returned and the extension is now red when active. The real fix for this is as mentioned in [Jeff Johnson's blog](https://lapcatsoftware.com/articles/blues.html) is to use a monochrome (usually just all black) transparent icon for the toolbar, this behaviour is seemingly completely undocumented by Apple. iOS also supports this behaviour, but since you cannot change the accent colour, it is always blue when active.

I figured since Safari already has some manifest overrides, the easiest way would be to override the toolbar icon using that. Below is how it looks during tested on macOS and iOS:

image

image

image

image


ajayyy:
nice, looks good

PR #1372, opened 2022-06-27
Implement Custom Channel Overrides

scaccoman:
This PR implements new custom channel override logic, which seems to be one of the most requested features:
- https://github.com/ajayyy/SponsorBlock/issues/435
- https://github.com/ajayyy/SponsorBlock/issues/1308
- https://github.com/ajayyy/SponsorBlock/issues/1251
- https://github.com/ajayyy/SponsorBlock/issues/1085
- https://github.com/ajayyy/SponsorBlock/issues/646

Screenshot 2022-06-27 at 17 47 54

Please see the attached screen recording for a brief demo:
https://user-images.githubusercontent.com/28684618/175990993-94e62021-bdc4-4309-960b-8730cd4bd097.mp4

There are some weird behaviours around `disabled` segment types, I'm not really sure how it's supposed to work. I would appreciate it if someone more familiar with the project (e.g. @ajayyy) could shed some light.

- [x] I agree to license my contribution under LGPL-3.0 **or** my contribution is from another project with a license compatible with LGPL-3.0

PR #2224, opened 2025-03-20
Finalized implementing channel specific settings

gosha305:

***

Combines implementation of channel specific category selections from #1372 (for the popup UI) and #1261. Thank you to both @zedseven and @scaccoman, as I likely wouldn't have even attempted this were it not for their previous work.

Additional minor changes also include:
 - Slightly modifying the popup UI in order to account for other languages having longer skip option and category names
 - Changing the `inherit` option's name to `global` in order to clearly convey that it uses the option selected in the global config
 - Adding a clarification of what the currently selected `global` option is, in parentheses
 
currently selected global option in parentheses

 - Adding a delete channel settings button in the popup (otherwise, the settings are only deleted if their category selections are empty)
 
delete channel settings button

 
Finally, removes the whitelist system entirely. Instead, existing whitelists are converted to have all non-disabled categories at `Show overlay` in the channel settings.

Also see https://github.com/ajayyy/ExtensionTranslations/pull/81

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,7 +41,7 @@
         "@types/wicg-mediasession": "^1.1.4",
         "@typescript-eslint/eslint-plugin": "^5.54.1",
         "@typescript-eslint/parser": "^5.54.1",
-        "chromedriver": "^110.0.0",
+        "chromedriver": "^134.0.0",
         "concurrently": "^7.6.0",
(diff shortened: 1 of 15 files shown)


ajayyy:
https://github.com/ajayyy/SponsorBlock/commit/40eabe6b1895042a5147c1812fb9372666b2df68

PR #2235, opened 2025-04-01
feat: Option to auto-skip non-music on youtube music only

FlorianZahn:

***

non-music segments are only auto-skipped on music.youtube.com. 
The option, that enables autoskipping all segments if the video is a music video is also affected by the option.

I also created pull requests for maze-utils and the language files repo. The commit already includes the updated commit reference for both repos.


ajayyy:
Ah, I see. I was pretty confused at what you meant by that, I'll make a fix

PR #2280, opened 2025-05-31
Hide skip sponsor button while YouTube's own ads are playing

pepijnmm:

***
What does the change do: SponsorBlock will now be hidden during YouTube's own ads.

The skip sponsor button can appear during YouTube ads. If you accidentally click skip, as sometimes YouTube ads start just before a sponsorBlock, YouTube will notify the user that it thinks that an adblocker is being used.

Example of what happened before the change:
image

### DIFF
--- a/public/content.css
+++ b/public/content.css
@@ -947,3 +947,8 @@ input::-webkit-inner-spin-button {
 .sponsorblock-chapter-visible {
 	display: inherit !important;
 }
+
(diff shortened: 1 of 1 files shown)


pepijnmm:
@ajayyy Can you tell me why you closed the pr?
For the people who use adBlock this change won't matter. but for the people who don't, this will make sure that they don't get the youtube premium warning
image

Issue #2272, opened 2025-05-20
Segments still loading


swanknight:
SB hasn't been working for me for a while now. I don't see any segments, and if I try to submit any segments, it just times out.
It just says "Segments still loading". Also, if I try to add a username, I get this error: "A connection error has occurred. Error code: -1".

PS: I'm using FF 128.10.1esr 

```
{
    "debug": {
        "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:128.0) Gecko/20100101 Firefox/128.0",
        "platform": "Win32",
        "language": "en-US",
        "extensionVersion": "5.12.1"
    },
    "config": {
        "autoSkipOnMusicVideosUpdate": true,
        "showZoomToFillError2": false,
        "showPopupDonationCount": 5,
        "hasShownYouCapNotice": true,
        "hideDiscordLaunches": 1,
        "autoSkipOnMusicVideos": false,
        "showChapterInfoMessage": false,
        "whitelistedChannels": 4,
        "payments": {
            "licenseKey": null,
            "lastCheck": 1666856597242,
            "lastFreeCheck": 1666541334836,
            "freeAccess": true,
            "chaptersAllowed": true
        },
        "forceChannelCheck": true,
        "categoryPillUpdate": true,
        "permissions": {
            "sponsor": true,
            "selfpromo": true,
            "exclusive_access": true,
            "interaction": true,
            "intro": true,
            "outro": true,
            "preview": true,
            "music_offtopic": true,
            "filler": true,
            "poi_highlight": true,
            "chapter": false
        },
        "categorySelections": [
            {
                "name": "sponsor",
                "option": 2
            },
            {
                "name": "intro",
                "option": 2
            },
            {
                "name": "outro",
                "option": 2
            },
            {
                "name": "interaction",
                "option": 2
            },
            {
                "name": "selfpromo",
                "option": 2
            },
            {
                "name": "music_offtopic",
                "option": 2
            },
            {
                "name": "preview",
                "option": 2
            },
            {
                "name": "chapter",
                "option": 0
            }
        ],
        "renderSegmentsAsChapters": true,
        "submissionCountSinceCategories": 60,
        "sponsorTimesContributed": 60,
        "chapterCategoryAdded": true,
        "showDeArrowPromotion": false,
        "disableSkipping": false,
        "minutesSaved": 3098.2362372234475,
        "skipCount": 5727,
        "invidiousInstances": 48,
[...]

PR #2319, opened 2025-08-19
Actually remove rejected promises from the pendingList

mini-bomba:

***
otherwise this may cause a failure to be "cached" and never removed, resulting in the inability to refresh segments until the page is refreshed

quickfix from the investigation [on discord](https://discord.com/channels/603643120093233162/603643256714297374/1407420707167670426)


mini-bomba:
quickly looking at the issue list, #2272 may be related
i did not experience the "-1" error directly, but I found that the thrown "error" that got stuck on the list had "status: -1"

PR #2336, opened 2025-09-19
Fix segment hiding/unhiding

mini-bomba:

***
[issue reported on discord](https://discord.com/channels/603643120093233162/603643256714297374/1418563576226119722)
The issue itself was triggered by [this line in PR #2334](https://github.com/ajayyy/SponsorBlock/pull/2334/files#diff-0885c230b87020856a81ffc8cf0303ab633e364e7f79071f60e6f0a3d8c26df7R91), as it shallow-cloned the segment object. The list item component relied on object mutation propagating up to the root popup component when unhiding, which was stopped in the list display due to this clone, and even worse - the changes would be discarded on the next rerender of the list.

This PR fixes this issue by making the list item trust it's own state variable for hidden status, which already existed but wasn't actually used.
This is safe, because:
- the state variable is initialized to the segment's value on first render, and will be updated when the segment's hidden property updates
- all uses of the SegmentListItem component set the segment's uuid as the key, ensuring that it will not get reused when segments update
- the root popup component will never derender the SegmentListComponent (no conditional rendering)
- components that use the SegmentListItem component will not derender it unless the segment lists change

this PR also uses the `React.useMemo` hook to cache segment nesting results

### DIFF
--- a/src/popup/SegmentListComponent.tsx
+++ b/src/popup/SegmentListComponent.tsx
@@ -28,8 +28,8 @@ enum SegmentListTab {
     Chapter
 }
 
-interface segmentWithNesting extends SponsorTime {
-    innerChapters?: (segmentWithNesting|SponsorTime)[];
+interface SegmentWithNesting extends SponsorTime {
+    innerChapters?: (SegmentWithNesting|SponsorTime)[];
 }
 
 export const SegmentListComponent = (props: SegmentListComponentProps) => {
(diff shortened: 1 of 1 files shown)


src/popup/SegmentListComponent.tsx
```diff
@@ -146,18 +150,32 @@ function SegmentListItem({ segment, videoID, currentTime, isVip, loopedChapter,
...
-        extraInfo = " (" + chrome.i18n.getMessage("hiddenDueToDownvote") + ")";
-    } else if (segment.hidden === SponsorHideType.MinimumDuration) {
-        // This one is too short
-        extraInfo = " (" + chrome.i18n.getMessage("hiddenDueToDuration") + ")";
-    } else if (segment.hidden === SponsorHideType.Hidden) {
-        extraInfo = " (" + chrome.i18n.getMessage("manuallyHidden") + ")";
+    // Update internal state if the hidden property of the segment changes
+    React.useEffect(() => {
+        setHidden(segment.hidden ?? SponsorHideType.Visible);
+    }, [segment.hidden])
```
ajayyy, line 159: Would this only be called once?

PR #2342, opened 2025-09-23
Rewrite advanced skip options parser to add "or" operator

mschae23:
Add `or` operator to advanced skip options, as well as `not` (which is syntactic sugar).

The new parser uses conventional precedence (`not` binds more tightly than `and`, which binds more tightly than `or`) and supports parentheses to override it. When converting the config back to text, parentheses will always be added when multiple operators are used, for clarity.

These changes do mean the advanced skip options language is now context-free instead of regular, so no more parsing with regex :)

I have not measured the performance changes in any way though, so I unfortunately couldn't tell you whether this is faster or slower than before.

~~Because I unfortunately do not know how to add a migration for it, previously saved skip options will fail to load on this version. For testing, you can run `browser.storage.local.set({"skipRules": []});` (back up your config before doing this!) in the `about:debugging` console for SponsorBlock. However, there should definitely be some kind of automatic fix before this can be used in production.~~

I have now written a migration for the storage format, which should hopefully™ be working.

The wiki page also requires changes to document the new features. For that, see <https://wiki.sponsor.ajay.app/w/Advanced_skip_options/Draft>.

---
I agree to license my contribution under GPLv3 (or any later version) and agree to allow distribution on app stores as outlined in [LICENSE-APPSTORE](https://github.com/ajayyy/SponsorBlock/blob/master/LICENSE-APPSTORE.txt)

To test this branch, follow the [instructions on the GitHub wiki that haven't been migrated to the real wiki yet for some reason](https://github.com/ajayyy/SponsorBlock/wiki/Testing-a-Pull-Request)

### DIFF
--- a/public/options/options.html
+++ b/public/options/options.html
@@ -329,6 +329,17 @@
 							</label>
 						</div>
 					</div>
+
+					<br />
+
+					<div data-type="selector" data-sync="segmentListDefaultTab">
+						<label class="optionLabel" for="segmentListDefaultTab">__MSG_segmentListDefaultTab__:</label>
+
+						<select id="segmentListDefaultTab" class="selector-element optionsSelector">
+							<option value="0">__MSG_SegmentsCap__</option>
+							<option value="1">__MSG_Chapters__</option>
+						</select>
+					</div>
 				</div>
 
 				<div data-type="toggle" data-sync="darkMode">
--- a/public/popup.css
+++ b/public/popup.css
@@ -591,6 +591,8 @@
 #sponsorTimesDonateContainer a {
   color: var(--sb-main-fg-color);
   text-decoration: none;
+
+  padding-left: 5px;
 }
 
 /*
--- a/src/components/SponsorTimeEditComponent.tsx
+++ b/src/components/SponsorTimeEditComponent.tsx
@@ -123,6 +123,8 @@ class SponsorTimeEditComponent extends React.Component<SponsorTimeEditProps, Spo
             style.marginTop = "15px";
         }
 
+        const borderColor = this.state.selectedCategory ? Config.config.barTypes[this.state.selectedCategory]?.color : null;
+
         // Create time display
         let timeDisplay: JSX.Element;
         const timeDisplayStyle: React.CSSProperties = {};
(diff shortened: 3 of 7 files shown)


src/utils/skipRule.ts
```diff
@@ -176,11 +195,678 @@ function isSkipRulePassing(segment: SponsorTime | VideoLabelsCacheData, rule: Ad
...
+type TokenType =
+    | "if" // Keywords
+    | "disabled" | "show overlay" | "manual skip" | "auto skip" // Skip option
+    | `${SkipRuleAttribute}` // Segment attributes
+    | `${SkipRuleOperator}` // Segment attribute operators
+    | "and" | "or" // Expression operators
+    | "(" | ")" | "comment" // Syntax
+    | "string" | "number" // Literal values
+    | "eof" | "error"; // Sentinel and special tokens
+
```
ajayyy, line 229: What are these tokens?

mschae23: `eof` is end of file (or end of input string, in this case), `error` is for any unexpected characters in the input

src/utils/skipRule.ts
```diff
@@ -176,11 +195,678 @@ function isSkipRulePassing(segment: SponsorTime | VideoLabelsCacheData, rule: Ad
...
+            c = peek();
+        }
+
+        return makeToken(error ? "error" : type);
+    }
+
+    let c = consume();
+
+    if (c === '"') {
+        // Parses string according to ECMA-404 2nd edition (JSON), section 9 “String”
```
ajayyy, line 452: Why is custom string and number parsing needed?

src/utils/skipRule.ts
mschae23: The lexer (or tokenizer, it's a synonym) converts the input string (list of characters) into a list of tokens (basically words) in `nextToken`, which are then used by the parser in `parseConfig` (actually, it converts it one token at a time, because that's how the parser uses it)

src/utils/skipRule.ts
ajayyy: ah yea, because it's per token. Makes sense

Issue #2362, opened 2025-10-17
Copy Public UserID button on YouTube pages doesn't work


yume-chan:
This one:

<img width="749" height="934" alt="Image" src="https://github.com/user-attachments/assets/937b03a7-5224-42a1-8d1a-ef9a2bbcf3e2" />

Console output:

```txt
[Violation] Permissions policy violation: The Clipboard API has been blocked because of a permissions policy applied to the current document. See https://crbug.com/414348233 for more details.
(anonymous)	@	YourWorkComponent.tsx:87
a	@	SegmentSubmissionComponent.tsx:65
Promise.then		
s	@	SegmentSubmissionComponent.tsx:65
(anonymous)	@	SegmentSubmissionComponent.tsx:65
o	@	SegmentSubmissionComponent.tsx:65
onClick	@	YourWorkComponent.tsx:86
Ue	@	react-dom.production.min.js:54
je	@	react-dom.production.min.js:54
(anonymous)	@	react-dom.production.min.js:55
Ro	@	react-dom.production.min.js:105
zo	@	react-dom.production.min.js:106
(anonymous)	@	react-dom.production.min.js:117
cu	@	react-dom.production.min.js:274
_e	@	react-dom.production.min.js:52
Wo	@	react-dom.production.min.js:109
Yt	@	react-dom.production.min.js:74
Wt	@	react-dom.production.min.js:73

Uncaught (in promise) NotAllowedError: Failed to execute 'writeText' on 'Clipboard': The Clipboard API has been blocked because of a permissions policy applied to the current document. See https://crbug.com/414348233 for more details.
    at YourWorkComponent.tsx:87:60
    at Generator.next (<anonymous>)
    at a (SegmentSubmissionComponent.tsx:65:6)
(anonymous)	@	YourWorkComponent.tsx:87
a	@	SegmentSubmissionComponent.tsx:65
Promise.then		
s	@	SegmentSubmissionComponent.tsx:65
(anonymous)	@	SegmentSubmissionComponent.tsx:65
o	@	SegmentSubmissionComponent.tsx:65
onClick	@	YourWorkComponent.tsx:86
Ue	@	react-dom.production.min.js:54
je	@	react-dom.production.min.js:54
(anonymous)	@	react-dom.production.min.js:55
Ro	@	react-dom.production.min.js:105
zo	@	react-dom.production.min.js:106
(anonymous)	@	react-dom.production.min.js:117
cu	@	react-dom.production.min.js:274
[...]

PR #2395, opened 2025-12-17
fix: Add missing clipboard write permission

ShreyMarwaha:
- [X] I agree to license my contribution under GPL-3.0 and agree to allow distribution on app stores as outlined in [LICENSE-APPSTORE](https://github.com/ajayyy/SponsorBlock/blob/master/LICENSE-APPSTORE.txt)

***

This fixes: https://github.com/ajayyy/SponsorBlock/issues/2362

Added 'clipboardWrite' to extension permissions in manifest.json and set 'allow' attribute to 'clipboard-write' on the info menu iframe to enable clipboard operations.

## Testing Results
Before:
image

After:
Copy button working as expected. No exception in console.
image

### DIFF
--- a/manifest/manifest.json
+++ b/manifest/manifest.json
@@ -17,8 +17,7 @@
     "permissions": [
         "storage",
         "scripting",
-        "unlimitedStorage",
-        "clipboardWrite"
+        "unlimitedStorage"
     ],
     "options_ui": {
         "page": "options/options.html",


ajayyy:
I don't think both are required. Just the iframe permission should work right?

ShreyMarwaha:
@ajayyy Yes, you are correct. Since we are writing to the clipboard on user interaction, we do not need the permission in the manifest.json. I have removed it.

Issue #2393, opened 2025-12-13
[SUGGESTION] Editing existing segments


djmrFunnyMan:
Sometimes a video has an existing segment but... it's too late or too early by a bit. RN the only way to correct this is to downvote to delete the segment and submit your own segment instead. If you don't do this, there're be two overlapping segments. This is quite tedious and annoying.

I see two potential solutions to fix this:

**First:** When you upload a segment that overlaps another of the same category, the two segments get merged into one segment. Ideally if your segment overlaps several previously seperate segments at once, it merges all of them (once again, IF they're of the same category)

**Second:** In the menu where you can view video segments (the one that has upvote, downvote, copy, hide, skip buttons) add an "edit" button. That allows you to manually edit it's start and end time, and perhaps some other things like change its category.

I don't know which option sounds more appealing. For me personally, I'd like to have _both_.

ajayyy:
There is a copy and downvote button right now under the pencil icon in the skip notice

jrial:
Adding my support for this feature request here. I notice a lot of people being confused about the correct segment category. An often made mistake is that people mark segments as sponsor segments, when they should have been unpaid/self promotion. Right now, all I can do about that is to downvote the original segment, create my own with the correct category, upload that and hope for the best. Comunity editing would hopefully alleviate this somewhat.

Apart from that, yes, a way to edit the timings would be nice too. Some people are quite imprecise or can't be bothered to put in the effort, and it ruins a ton of transitions.

ajayyy:
@jrial click the pencil button and choose "change category"

jrial:
@ajayyy Ah, I see, in the little toast there's indeed a pencil icon for editing the submission. However, that's not really streamlined for mass edits. I was expecting to find it in the info screen. There I can do the same things as from the toast: upvote, downvote, copy segment ID. And some extra stuff I can't do from the toast: preview and skip. But not edit.

<img width="453" height="394" alt="Image" src="https://github.com/user-attachments/assets/ba66f623-adbb-4ff7-bad4-071c02eae52e" />

Rationale: I've been noticing people miscategorising unpaid promotions as sponsor segments on tens of videos from the same channel. It's tedious to advance them all to the point right before the segment and wait for the toast so I can edit. I would love to correct all these miscategorisations, if I can do it without wasting an entire evening. I can imagine I'm not the only one in a similar situation.

PR #2422, opened 2026-02-06
Implemented editing currents segments

pPyrius:

Implemented new edit segment feature 

image

Trying to solve issue #2393


ajayyy:
What does it do when you edit them

pPyrius:
Basically it create a new segment and dislike the previous one

PR #2501, opened 2026-06-08
Fix #2489

### LINKED ISSUE #2489: Cannot create segments throught YT embeds (TypeError: can't access property "startsWith", r is undefined)
Using r3dfox 150.0.3/Firefox 150 & latest SB beta. Every time I press the Mark segment start button on a YT embed video it doesn't do anything and when I check in the console I notice that it throws me this error "TypeError: can't access property "startsWith", r is undefined" every time I click on the button.

https://github.com/user-attachments/assets/954c935f-12e2-40c1-a056-dffa6b2c2472

This error doesn't happen on normal YT though.

mattfbacon:

***

`playButtonSVGDataNew` can be `undefined` which causes `startsWith` to throw an exception. I assume Typescript is not configured with strict null checking so this wasn't caught.


ajayyy:
Thanks

PR #2517, opened 2026-07-13
Fix hidden state changes from minimum duration

ShiroKSH:
## Summary

- preserve downvoted and manually hidden segments when applying the minimum-duration filter
- restore only segments hidden by that filter when the threshold changes or is removed
- apply `minDuration` updates to the current video immediately

## Root cause

The minimum-duration predicate treated the `MinimumDuration` enum member as an unconditional truthy value, so it relabeled every short segment. The previous zero-threshold branch also left filter-hidden segments unchanged.

## Validation

- `npm run test-without-building -- --runInBand`
- `npm run lint`
- `npm run build:chrome`
- `npm audit --omit=dev --json` (0 production vulnerabilities)

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -371,6 +371,7 @@ function contentConfigUpdateListener(changes: StorageChangesObject) {
                 updateVisibilityOfPlayerControlsButton()
                 break;
             case "categorySelections":
+            case "minDuration":
                 channelIDChange();
                 break;
             case "barTypes":
--- a/src/utils/skipProfiles.ts
+++ b/src/utils/skipProfiles.ts
@@ -85,14 +85,13 @@ function getSkipProfileValue<T>(key: keyof CustomConfiguration): T {
 export function hideTooShortSegments(sponsorTimes: SponsorTime[]) {
     const minDuration = getSkipProfileNum("minDuration");
 
-    if (minDuration !== 0) {
-        for (const segment of sponsorTimes) {
-            const duration = segment.segment[1] - segment.segment[0];
-            if (duration > 0 && duration < minDuration && (segment.hidden === SponsorHideType.Visible || SponsorHideType.MinimumDuration)) {
-                segment.hidden = SponsorHideType.MinimumDuration;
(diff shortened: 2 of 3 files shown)


ajayyy:
What is the goal of this pull request

ShiroKSH:
The goal is to keep the minimum-duration filter from overwriting intentional hidden states.

A short segment marked `Downvoted` or manually `Hidden` was relabeled as `MinimumDuration`; after changing or removing the threshold, it could become visible again. This PR limits that temporary transition to `Visible` ↔ `MinimumDuration` in `src/utils/skipProfiles.ts:90`, applies `minDuration` changes to the current video in `src/content.ts:374`, and adds regression coverage for both cases in `test/skipProfiles.test.ts`.

ajayyy:
What is the user-level change you are trying to achieve

What are the steps to reproduce this bug in the UI, with screenshots

ShiroKSH:
This came from code review rather than a user report. The user-level goal is that an explicit "Hide segment" choice remains authoritative when a per-tab skip profile changes its minimum-duration setting.

UI reproduction on the current code:

1. Open a video with a segment shorter than 10 seconds.
2. In the extension popup's segment list, use the eye button to mark that segment as **Hide segment**.
3. Create a skip profile with **Minimum duration (seconds)** set to `10`, then apply it **For this tab** from the popup's skip-profile menu.
4. Switch the same tab to a profile with the minimum duration set to `0` (or below the segment's duration).

Before this PR, the explicitly hidden segment is first relabeled as `hidden: too short`, then becomes visible again when the profile changes. The expected result is that it remains manually hidden; minimum duration should only toggle segments that it hid itself.

I do not have an authentic UI screenshot to attach, so I will not fabricate one. The regression test covers the same state transition in `test/skipProfiles.test.ts`. If this profile-switch edge case is not sufficient reason for a change, I am happy to close the PR.

ajayyy:
Hi,

I have reported your spamming and bot activity to github

ShiroKSH:
> Hi,
> 
> I have reported your spamming and bot activity to github

hd, np

